PR #3, opened 2015-07-26
skip build if resulting .tar.gz already exists

ralphtheninja:


ralphtheninja:
A side effect of this is if you do

```
$ prebuild -t 1.8.1 -t 1.8.2
```

That will only build *one* `.tar.gz` since it would result in the same abi

mafintosh:
Great!

mafintosh:
Released as 1.2.0. I've added a `--force` flag that always builds as well

PR #7, opened 2015-07-29
use mv instead of fs.rename to get around EXDEV errors

ralphtheninja:
On debian systems you may get the `EXDEV` error when moving files between devices, e.g.

exdev


mafintosh:
Actually reading the `mv` README this kinda defeats the purpose of writing to a tmp file in the first place as mv will fallback to doing a copy. We rename because of atomicity. I changed to just write the tmp file next to the cache file instead. That should fix the EXDEV as well.

ralphtheninja:
Cool. Even better.

PR #9, opened 2015-07-30
more elaborate downloading

ralphtheninja:
Download from github (default)

```
$ prebuild --download
```

Download from other host (expansion works)

```
$ prebuild --download https://example.com/{name}-v{version}-{abi}-{arch}.tar.gz
```

It's also possible to download based on the `binary` property in package.json

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -84,8 +84,7 @@ prebuild(targets.shift(), function done (err, result) {
 })
 
 function downloadPrebuild () {
-  var name = pkg.name + '-v' + pkg.version + '-node-v' + process.versions.modules + '-' + rc.platform + '-' + rc.arch + '.tar.gz'
-  var url = github(pkg) + '/releases/download/v' + pkg.version + '/' + name
+  var url = getDownloadUrl()
   var cache = path.join(NPM_CACHE, url.replace(/[^a-zA-Z0-9.]+/g, '-'))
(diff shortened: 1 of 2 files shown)


ralphtheninja:
@mafintosh Let me know if this looks ok or if we should adjust some stuff. Once we're done with that, I can update `README`

ralphtheninja:
See this as reference https://github.com/mapbox/node-pre-gyp#versioning

bin.js
```diff
@@ -116,6 +116,46 @@ function downloadPrebuild () {
...
+    Object.keys(values).forEach(function (key) {
+      var regexp = new RegExp('\{' + key + '\}', 'g')
+      template = template.replace(regexp, values[key])
+    })
+    return template
+  }
+
+  function urlTemplate () {
+    if (typeof rc.download == 'string') return rc.download
+    var package_name = '{name}-v{version}-node-v{abi}-{platform}-{arch}.tar.gz'
```
mafintosh: could you camelCase this?

ralphtheninja: Absolutely

bin.js
```diff
@@ -97,6 +96,7 @@ function downloadPrebuild () {
     var status = 0
 
     log.http('request', 'GET ' + url)
+
```
mafintosh: could you remove the white space here?

mafintosh:
Released in 2.3.0 - feel free to add some documentation on how this works

ralphtheninja:
:+1: Will do!

PR #15, opened 2015-08-01
download tests

ralphtheninja:
So after some poking around I ended up with this. I found a couple of bugs. See comments below.


ralphtheninja:
@mafintosh Maybe after some adjustments we could squash it once we're content?

mafintosh:
:+1: for squashing

download.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    pump(fs.createReadStream(cachedPrebuild), zlib.createGunzip(), tfs.extract(rc.cwd, {readable: true, writable: true}).on('entry', updateName), function (err) {
+      if (err) return cb(err)
+      if (binaryName) {
+        try {
+          var resolved = path.resolve(rc.cwd, binaryName)
+          require(resolved)
+          log.info('unpack', 'required ' + resolved + ' successfully')
+          return cb()
```
mafintosh: don't try-catch the cb() call. we might end up calling cb twice when doing this

test/util-test.js
```diff
@@ -0,0 +1,165 @@
+var test = require('tape')
+var util = require('../util')
+var fs = require('fs')
+
+test('prebuildCache() ends with _prebuilds folder', function (t) {
+  process.env.APPDATA = 'somepathhere'
+  t.ok(util.prebuildCache().indexOf('/npm-cache/_prebuilds') > 0, 'APPDATA set')
```
mafintosh: shouldn't this check if it contains APPDATA?

util.js
```diff
@@ -0,0 +1,69 @@
...
+  })
+}
+
+function expandTemplate (template, values) {
+  Object.keys(values).forEach(function (key) {
+    var regexp = new RegExp('\{' + key + '\}', 'g')
+    template = template.replace(regexp, values[key])
+  })
+  return template
+}
```
mafintosh: you should put this one in a module

download.js
ralphtheninja: Good catch!

test/util-test.js
ralphtheninja: Yes it should! :)

util.js
ralphtheninja: Good idea. I'll do that.

download.js
ralphtheninja: Updated.

test/util-test.js
ralphtheninja: Updated.

ralphtheninja:
@mafintosh Ready for squashing?

mafintosh:
@ralphtheninja looks good! could you add a test script to npm as well? then i'll add travis afterwards

mafintosh:
lol script is already there. ready for squashing :tada:

mafintosh:
2.3.2

PR #18, opened 2015-08-03
Break out all the things

ralphtheninja:
Let me know about adjustments. Upload is tested now. Will obviously squash later.


ralphtheninja:
Added tests for configuration (and flags on cli). However, tests started to fail on iojs 3.0.0. Not really sure what's going on here. The tests work fine if you switch to e.g. iojs 2.5.0

PR #28, opened 2015-08-31
add support for --strip + moar tests

ralphtheninja:
The command line flags used for `strip(1)` comes from https://github.com/frida/frida/blob/master/releng/release.py#L77-L82

Closes https://github.com/mafintosh/prebuild/issues/10

cc @oleavr


ralphtheninja:
@mafintosh would be sweet if you could try this out. I got the following results on `leveldown`:

prebuild-stripped

oleavr:
@ralphtheninja Sweet! Presumably other UNIXes would either follow the GNU (`-s` / `--strip-all`) or the Apple/BSD (`-Sx`) style, although I haven't looked into that. As for Windows, when using the Microsoft toolchain, the symbols are usually saved to a `.pdb` file so that makes things quite simple.

ralphtheninja:
@oleavr Cool. Well, we can add those flags later whenever we need them. Or maybe someone else comes along with a PR :)

ralphtheninja:
And yes, having symbols separately is something M$ handles well :)

ralphtheninja:
2.4.0

PR #30, opened 2015-09-04
enable building with --debug

ralphtheninja:
* make sure prebuild catches --debug from npm environment (npm i --debug)
* if --debug is set, build to build/Debug (util.releaseFolder()) + tests
* pass in --debug to node-gyp if set
* factor out gyp code + tests

Closes https://github.com/mafintosh/prebuild/issues/29

### DIFF
--- a/build.js
+++ b/build.js
@@ -1,12 +1,11 @@
 var path = require('path')
 var fs = require('fs')
 var install = require('node-gyp-install')
-var nodeGyp = require('node-gyp')()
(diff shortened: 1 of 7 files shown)


mafintosh:
:+1:

ralphtheninja:
New patch or new minor? Technically we were using `--debug` before this but only for the download url.

mafintosh:
Minor

ralphtheninja:
2.5.0

PR #31, opened 2015-09-06
add better error message when no prebuilts are available

### LINKED ISSUE #19: improve/add message when prebuild not available and falling back to build
in the case where a prebuilt binary is not available, it would be awesome if a message like this was printed out to the user to help them understand what is happening:

```
Prebuilt binaries for your <node or iojs> version, <version number>, are not available for download.
We will now try to compile from source.
```

We have had some people get confused when their compile failed (e.g. they were using iojs 3.0.0 and prebuilts werent available and the compile failed). It would be nice if the output was super clear so they can understand whats goin on

ralphtheninja:
Closes #19

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -42,7 +42,7 @@ if (rc.download) {
   return download({pkg: pkg, rc: rc, log: log}, function (err) {
     if (err) {
       log.warn('install', err.message)
-      log.info('install', 'Falling back to compilation')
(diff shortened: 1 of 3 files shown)


ralphtheninja:
@mafintosh Any objections?

mafintosh:
No looks good!

PR #32, opened 2015-09-06
properly report how many files were uploaded

ralphtheninja:
I noticed that prebuild doesn't correctly report how many prebuilts that were actually uploaded, e.g. if we are trying to upload 6, it will report that 6 were uploaded even though they were filtered out. Also, while fixing this, why not also print out the prebuilts that were found?

uploaded-releases


mafintosh:
:+1:

ralphtheninja:
2.5.2

PR #37, opened 2015-09-20
add support for @scoped/module-names

ralphtheninja:
Closes https://github.com/mafintosh/prebuild/issues/36

cc @corbinu

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "prebuild",
-  "version": "2.5.2",
+  "version": "2.5.3",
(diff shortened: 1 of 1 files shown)


mafintosh:
This would make @test/module and test-module clash though right?

corbinu:
This looks good to me.

@mafintosh maybe a separate feature is in order to allow "prebuild_name" in the package.json which if present allows the normal package name to be overridden?

ralphtheninja:
@mafintosh Aah yes it would. I'll update today.

ralphtheninja:
I'll create a new PR based on top of https://github.com/mafintosh/prebuild/pull/39

PR #41, opened 2015-09-22
Make --download search local prebuilds folder first.

brett19:
This allows me to provide all the prebuilds on my build
server instead of having to rebuild the modules
when not neccessary.

### DIFF
--- a/download.js
+++ b/download.js
@@ -9,40 +9,57 @@ var util = require('./util')
 function downloadPrebuild (opts, cb) {
   var downloadUrl = util.getDownloadUrl(opts)
   var cachedPrebuild = util.cachedPrebuild(downloadUrl)
+  var localPrebuild = util.localPrebuild(downloadUrl)
   var tempFile = util.tempFile(cachedPrebuild)
 
   var rc = opts.rc
   var log = opts.log
 
-  fs.exists(cachedPrebuild, function (exists) {
-    if (exists) return unpack()
+  if (opts.nolocal) return download()
 
-    log.http('request', 'GET ' + downloadUrl)
-    var req = get(downloadUrl, function (err, res) {
-      if (err) return onerror(err)
-      log.http(res.statusCode, downloadUrl)
-      if (res.statusCode !== 200) return onerror()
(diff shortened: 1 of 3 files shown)


brett19:
Might be getting a little carried away on the releases :D

brett19:
Oh fantastic, this change REALLY blew up the tests.

brett19:
Interesting.  Just noticed `npm test` fails with any node version in the 4.x.y range.

ralphtheninja:
@brett19 I think that's related to `a-native-module` that doesn't have any prebuilts for node 4.0.0 yet. (using that module for tests)

ralphtheninja:
Just guessing now, might be unrelated :)

brett19:
Yea, I just mentioned it, but didn't bother looking at why it was happening at all. :)

brett19:
Any thoughts on this PR?  It helps slightly, but I'm worried it may cause some weird issue.  As an aside as well, I think we should make `--download` be `nolocal:true`, but also include a `--install` for use in `package.json>scripts>install` which searches locally as well.

ralphtheninja:
I'm not sure of the use case here to be honest. You say you have a build server? Why can't you just put the binaries there and download from it?

brett19:
Jenkins, and that makes it a pain in the butt to download from.  Could possibly work though, seems a little silly to do http download of a file that already exists in the directory tree though (under /prebuilds).

ralphtheninja:
But if you have published the binaries, they will be downloaded from github and also cached locally in `~/.npm/_prebuilds`

ralphtheninja:
On second thought. You obviously build the binaries on jenkins so it will be an unnecessary dowload from github the first time.

brett19:
I also want to run the tests before I upload to GitHub.

ralphtheninja:
@brett19 *nods* It's a good use case then imo :)

brett19:
PR hell.  Can we merge these sometime soon :+1:

brett19:
This should be good to merge I think?

PR #43, opened 2015-09-23
Scoped modules

ralphtheninja:
Lets try this again.

@mafintosh Your idea with doing mkdirp made things so much easier. After all, we are only interested in the actual `.tar.gz` file when packing the file and uploading it. Also, while downloading we don't care about the scope, the github repo is still unique.


mafintosh:
:+1:

PR #44, opened 2015-09-25
Fix missing npm dir

brett19:


brett19:
Multiple active pull requests is driving me crazy...

PR #49, opened 2015-09-26
Read proxy

brett19:


ralphtheninja:
@brett19 Does this PR solve https://github.com/mafintosh/prebuild/issues/35 ?

brett19:
It does not, it simply makes http-proxy and https-proxy available through RC, which includes importing it from npm config or your environment.  Actually tying this in to simple-get or replacing simple-get with something that handles proxies by default still needs to be completed.

ralphtheninja:
> Actually tying this in to simple-get or replacing simple-get with something that handles proxies by default still needs to be completed.

Gotcha. It feels pretty urgent. I don't think that most people use proxies but some have had problems with it.

brett19:
I originally tried to replace our use of simple-get with the request module, but that failed miserably due to the way that request implicitly reads the whole request body instead of allowing us to read manually.  This meant I had to pipe the data directly to a file, which then broke a bunch of 'failed request should not leave temporary files' tests.  I'm sure it would be easy to solve if someone sat down for a few minutes and really tried, but I was in a rush to finish other non-prebuild things that day so I opted to just contribute the rc stuff, and let someone else work out the `download.js` details of it.
Cheers, Brett

PR #54, opened 2015-10-04
readme tweaks

ralphtheninja:


ralphtheninja:
Holding off with merge. Will remove the hack.

PR #55, opened 2015-10-11
move require test out from download

ralphtheninja:


ralphtheninja:
Gah I can't make my mind up. Closing again :(

PR #57, opened 2015-11-26
build all targets with --all

ralphtheninja:
@mafintosh Do you think we should really support node 0.8 as well? It definitely doesn't build for `leveldown`. Not exactly sure why, but I suspect that `nan` doesn't support it.


ralphtheninja:
@mafintosh Hmm, why isn't travis building?

mafintosh:
@ralphtheninja i'm fine with excluding 0.8 from --all

mafintosh:
looking good otherwise

ralphtheninja:
> @ralphtheninja i'm fine with excluding 0.8 from --all

Cool. I'll clean up PR and merge.

ralphtheninja:
2.7.0

PR #59, opened 2015-12-24
Moar logging during download

ralphtheninja:
Also changed the tests a bit. They were testing too much internal details imo. I don't think we should bump major.


ralphtheninja:
2.7.2

PR #60, opened 2015-12-27
Added --install option which will download only when installed via npm.

brett19:


ralphtheninja:
:+1:

ralphtheninja:
2.8.0

ralphtheninja:
@brett19 Thanks for this btw. It's much better to do `--install`!

PR #62, opened 2015-12-29
update readme with --install changes

ralphtheninja:
Maybe we should encourage people to use `--install` rather than `--download`?

Basically renamed all things from `--download` to `--install` and moved that up a bit, before the upload section.


ralphtheninja:
@brett19 @mafintosh Let me know what you think

brett19:
I think that `--download` is still useful in some cases, but the default should probably indeed be `--install`.  You should make sure to clarify what the difference between `--install` and `--download` is in the README though.  I didn't see that when I glanced over it.

ralphtheninja:
2.8.0

PR #68, opened 2016-01-14
add --abi option

juliangruber:
My end goal is to have seamless electron integration, but for now this should enable a more manual process for downloading the desired prebuild

### DIFF
--- a/download.js
+++ b/download.js
@@ -81,15 +81,26 @@ function downloadPrebuild (opts, cb) {
     pump(fs.createReadStream(cachedPrebuild), zlib.createGunzip(), tfs.extract(rc.path, {readable: true, writable: true}).on('entry', updateName), function (err) {
       if (err) return cb(err)
       if (!binaryName) return cb(new Error('Missing .node file in archive'))
+
       var resolved
       try {
         resolved = path.resolve(rc.path || '.', binaryName)
-        require(resolved)
       } catch (err) {
         return cb(err)
       }
-      log.info('unpack', 'required ' + resolved + ' successfully')
-      cb(null, resolved)
+      log.info('unpack', 'resolved to ' + resolved)
+
+      if (rc.abi === process.versions.modules) {
+        try {
+          require(resolved)
+        } catch (err) {
+          return cb(err)
+        }
+        log.info('unpack', 'required ' + resolved + ' successfully')
+        cb(null, resolved)
+      } else {
+        cb(null, resolved)
+      }
     })
   }
 
--- a/util.js
+++ b/util.js
@@ -16,7 +16,7 @@ function getDownloadUrl (opts) {
     patch: opts.pkg.version.split('.')[2],
     prerelease: opts.pkg.version.split('-')[1],
     build: opts.pkg.version.split('+')[1],
-    abi: process.versions.modules,
+    abi: opts.rc.abi || process.versions.modules,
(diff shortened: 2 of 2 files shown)


mafintosh:
:+1: @ralphtheninja ?

brett19:
I'm -1 to this change as it can cause confusion in non-electron cases.  Why not simply add a temporary --electron option for now?

juliangruber:
@brett19 reaching `--electron` is the goal, i wrote this in the process of getting there and thought it would still be useful.

about non-electron cases, i think we just need to make clear which commands this option applies to.

ralphtheninja:
@juliangruber if you want we could create a separate issue for this and spec out what we need, I can help out as well

ralphtheninja:
Hmm @brett19 is on to something, but `--abi` is useful too.

As I see it, we want to

1. install prebuilt binaries from github, based on the abi version of electron, so we need the abi from somewhere, right?
2. if we can't find any prebuilt binaries, we want to call out to `node-gyp` (like we do now) but with some different settings, e.g. changing `HOME` and passing on `--dist-url`:

`HOME=~/.electron-gyp node-gyp rebuild --target=$TARGET --arch=$ARCH --dist-url=https://atom.io/download/atom-shell`

taken from [electron-demo](from here https://github.com/Level/electron-demo/blob/master/package.json#L9)

So first of all, to solve 1. we need the abi somehow and the question is how we get that. There are different methods for this, one being the one @juliangruber just mentioned [here](https://github.com/mafintosh/prebuild/issues/61#issuecomment-171641418), another being using either a hardcoded `--target=0.X.Y` or by using https://github.com/ralphtheninja/electron-version and using `node-gyp` to actually download the headers for electron with that target and using https://github.com/mafintosh/prebuild/blob/master/abi.js#L4 to just parse it out from the headers.

Did that make sense?

ralphtheninja:
And btw, we can just use this PR to discuss the code and update. Once we're good, we just squash and clean up. Sound good?

juliangruber:
working on the real `--electron` thing now

juliangruber:
I've been working on a new module, https://github.com/juliangruber/require-recompile, that tackles this problem with a new approach: It patches `require()` so when you `require()` a module that has been built for a different version of node, it will recompile it. It will either call `node-gyp rebuild` or shell out the `prebuilt`, but for that `prebuilt` needs to support the `--abi` option.

So, can I get another +1 or a merge for this?

brett19:
Hey Guys,
I guess what I'm confused about here is: does node-gyp support building for an ABI that is not the current one?  I was under the impression that it could not, but maybe it was added more recently or I was simply unaware.
Cheers, Brett

juliangruber:
@brett19 yes it's possible: `node-gyp rebuild --target=X`

ralphtheninja:
+1 You can just merge https://github.com/mafintosh/prebuild/pull/70 which will close this one

juliangruber:
@ralphtheninja i'm just gonna merge this one since i suspect the require-recompile way will have way less edge cases than `--electron`, for example that it works on the whole dependency tree.

ralphtheninja:
@juliangruber Are you fine using `github:mafintosh/prebuild` in `require-rebuild` for now?

juliangruber:
yup no worries, it's still experimental

ralphtheninja:
2.9.0

PR #70, opened 2016-01-15
Add --electron

juliangruber:
approach taken:
- write a function that transforms the `opts` object in case `opts.rc.electron` is truthy
- ask local electron (from electron-prebuilt) or global electron about its abi
- set `opts.rc.abi` accordingly

questions:
- the script written to `/tmp/electron_abi.js` could also just live as a static file in this module
- is this enough, or should we ensure electron can require this module and if not recompile with the correct settings? (i guess this can just go into the next release)

### DIFF
--- a/download.js
+++ b/download.js
@@ -13,33 +13,24 @@ function transformElectron (opts, cb) {
   if (!opts.rc.electron) return process.nextTick(cb)
 
   var log = opts.log
-  var local = './node_modules/.bin/electron'
-  fs.stat(local, function (err) {
-    var electron = err
-      ? 'electron'
-      : local
+  log.info('fetching electron abi')
 
-    log.info('fetching electron abi from "' + electron + '"')
-    ask(electron)
-  })
+  var script = '/tmp/electron_abi.js'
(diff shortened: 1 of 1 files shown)


ralphtheninja:
@juliangruber kill this or keep it?

juliangruber:
i'd say let's leave it around

ralphtheninja:
K

hintjens:
@juliangruber can you resolve the merge conflicts and push again?

juliangruber:
rebased and pushed. fwiw i'm using https://github.com/juliangruber/require-rebuild now and it's working really well so far, so i personally am not dependent on this option any more. i still think it would be great to have though.

octalmage:
I would love to have a solution for this that works out of the box. I'm running into an issue with this now. electron-rebuild just downloads the regular node build and running with Electron gets the module mismatch error. I'd rather use this then node-pre-gyp.

ralphtheninja:
@juliangruber Maybe close this in favor of https://github.com/mafintosh/prebuild/pull/141 ?

juliangruber:
yep, #141 is better

PR #76, opened 2016-01-20
pass --target to build()

juliangruber:
so `prebuild download --target=5.1.1` builds for `5.1.1` if it has to build. it still defaults to `process.version`


ralphtheninja:
Good catch!

ralphtheninja:
2.9.0

PR #79, opened 2016-01-27
use os-homedir rather than home-dir

ralphtheninja:
`os-homedir` ponyfills `require('os').homedir`, see https://github.com/sindresorhus/os-homedir/blob/master/index.js#L24

I prefer to use `require('os').homedir` if it's available


ralphtheninja:
@mafintosh ping

juliangruber:
+1

PR #86, opened 2016-02-11
tweak README

ralphtheninja:
get rid of ABI section and unify that with Build, it was confusing

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # prebuild
 
-A command line tool for easily doing prebuilds for multiple version of node/iojs on a specific platform
+A command line tool for easily doing prebuilds for multiple version of node/iojs on a specific platform.
 
 ```
 $ npm install -g prebuild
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,13 +18,32 @@ $ npm install -g prebuild
...
+
+Optionally, to always build for the above versions you can add the following to `~/.prebuildrc`.
+
+``` ini
+prebuild[] = 0.10.42
+prebuild[] = 0.12.10
+prebuild[] = 4.3.0
 ```
 
+Note that `~/.prebuildrc` is a global configuration, i.e. when building *any* native module on your system. If this is not what you want you should consider adding a `.prebuildrc` to your project. This way the module determines which version it supports rather than a global setting.
```
juliangruber: should we make the local `.prebuildrc` case the default?

ralphtheninja: Yes! Good call :) Maybe not even mention the global setting?

juliangruber: sounds good to me!

ralphtheninja: We could maybe mention that you can use `~/.prebuildrc` for storing your token, which is a setting you really want globally

juliangruber: true, i think that's the common use case

ralphtheninja:
I'll squash and merge on +1

mafintosh:
Looks good to me

ralphtheninja:
3.0.1

PR #91, opened 2016-02-14
Errors

ralphtheninja:
Main reason for this was to get more detailed error information when no prebuilts can be found (see https://github.com/Level/leveldown/issues/254). That error just tells us we couldn't find any prebuilts for a certain node version. This way we don't need to ask "hey what platform, arch are you using?", we already have that information in the error.

Also moved out errors into `error.js`, which also makes it easier to test.

### DIFF
--- a/abi.js
+++ b/abi.js
@@ -1,5 +1,6 @@
 var gypinstall = require('./gypinstall')
 var util = require('./util')
+var error = require('./error')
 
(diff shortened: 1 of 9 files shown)


juliangruber:
+1 looks tidier as well!

ralphtheninja:
3.0.3

PR #93, opened 2016-02-16
Added --backend=node-ninja option

hintjens:
This lets us use node-ninja as backend instead of node-gyp. There
are two main reasons to want this:

- node-ninja fixes a bug in node-gyp/gyp where makefiles end up
  in arbitrary places on the filesystem (while the bug may get
  fixed in gyp, this is a blocking issue.)

- node-ninja offers a --builddir option for compiler output; we
  can use this to make incremental builds by puttting each ABI
  version in its own build directory.

More generally this removes the long-term dependency on gyp, as
we can add arbitrary backends. Node-ninja will aim to support the
newer Ninja make system instead of gyp/make.

Caveat: I did not make the backends optional dependencies, so you
both node-gyp and node-ninja are marked as dependencies.

### DIFF
--- a/abi.js
+++ b/abi.js
@@ -22,9 +22,9 @@ function getAbi (opts, version, cb) {
   })
 
   function tryReadFiles (readCb) {
-    util.readGypFile(version, 'node_version.h', function (err, a) {
+    util.readGypFile(opts, version, 'node_version.h', function (err, a) {
       if (err) return readCb(err)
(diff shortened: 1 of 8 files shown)


mafintosh:
@ralphtheninja can you spot why this is failing?

ralphtheninja:
@hintjens Merged your commit into `node-ninja` branch https://github.com/mafintosh/prebuild/tree/node-ninja

PR #95, opened 2016-02-16
Remove iojs from --all

ralphtheninja:


ralphtheninja:
I added last commit directly onto previous PR, so we can just merge this and the other will close.

ralphtheninja:
I like cleaning up remote branches as well. Does anyone have a problem with that?

hintjens:
In our projects we don't use branches at all. Merge to master and when/if it breaks, fix it. That tends to focus attention and keep things flowing. It works best when commits are minimal and focused.

ralphtheninja:
> In our projects we don't use branches at all.

Meaning you always use forks, right?

ralphtheninja:
4.0.0

hintjens:
@ralphtheninja not even, any more. We release from master, with various techniques to never break user space.

ralphtheninja:
> Merge to master and when/if it breaks, fix it

But what do you merge from if you don't have any forks and no remote branches? This is intriguing :)

hintjens:
Sorry, sorry, that kind of fork, I understand... everyone of course works in their own fork. That means we don't make projects in personal accounts, always in an organization. Everyone does fork/commit/push/pull request back to master.

I've not experimented much with using the GitHub UI to work directly on master and isolate commits in automatic branches. Perhaps for non-coders.

The goal of this is to avoid branches on the main repo, because they add complexity (to the repo, and to using git) and they increase change latency (how long before a change is proven innocent or guilty... minutes, or days...)

ralphtheninja:
> Sorry, sorry, that kind of fork, I understand... everyone of course works in their own fork. That means we don't make projects in personal accounts, always in an organization. Everyone does fork/commit/push/pull request back to master.

Makes sense. This is my preferred workflow.

PR #98, opened 2016-02-16
Fixed inconsistencies in the help text

hintjens:
We still had the -pb option in help.txt, and --backend wasn't in
README.md. They're consistent now.


ralphtheninja:
good catch

ralphtheninja:
4.0.1

PR #99, opened 2016-02-16
Problem: Failed to locate `node.h` and `node_version.h`

hintjens:
Hits when installing new ABI versions. Cause was gypinstall.js not
passing the backend down to gyp, so headers were installed into
$HOME/.node-gyp when using the --backend=node-ninja.

Solution: gypinstall.js passes opts.backend down when calling backend.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,8 @@ This command would:
 * Create a release on GitHub, if needed
 * Upload all binaries to that release, if not already uploaded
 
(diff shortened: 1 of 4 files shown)


ralphtheninja:
Apart from `standard`. LGTM.

hintjens:
OK, fixed those two issues. Thanks for the pointers.

ralphtheninja:
:+1:

ralphtheninja:
4.0.2

PR #102, opened 2016-02-26
Accept an environment variable for the token

rgbkrk:
This allows us to do the uploads in a Docker container by doing a build and run with `-e GITHUB_TOKEN=TOKEN`


rgbkrk:
Updated, thanks!

ralphtheninja:
You should already be able to do this since we use [rc](https://github.com/dominictarr/rc), by doing `-e prebuild_upload=TOKEN`

rgbkrk:
Ah cool, I'll do that instead. Thanks.

PR #111, opened 2016-04-02
Removed check for invalid archive on download

piranna:
I've tried to implement the checking for uploading, but implementing it on the binary make it impossible to test, and doing so on `upload.js` file makes all test to crash because fixtures are not correctly generated (don't have a `.node` file inside), so I think the best option is to remove any checking at all, leading to other use cases besides prebuild Node.js compiled modules like allowing to have prebuild binaries of programs like PhantomJS or my personal use case (gcc and binutils).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "prebuild",
-  "version": "5.1.1",
+  "version": "5.1.2",
   "description": "A command line tool for easily doing prebuilds for multiple versions of node/iojs or electron on a specific platform",
   "scripts": {
     "test": "tape test && npm run audit && npm run lint",


ralphtheninja:
@piranna Mind rebasing this onto latest master?

piranna:
Rebased.

ralphtheninja:
Can you tell us a little more about your use case and what you're trying to achieve? I can update the tests if you like.

piranna:
I'm trying to pack some binaries needed to build NodeOS, like the GCC-based cross-compiler or QEmu emulator. Due to this, I'm generating the tarball by hand since I don't have a .node file for it, and so I don't have it when downloading it so the check will always fail. Maybe this would be included on the new `prebuild-install` module?

ralphtheninja:
> Maybe this would be included on the new prebuild-install module?

That's a good idea! I guess once `prebuild-install` is operational, we will clean up this module and release a new major.

piranna:
Good to know :-) Do you want that I cherry-pick this change and set it on the prebuild-install repo?

ralphtheninja:
> Do you want that I cherry-pick this change and set it on the prebuild-install repo?

+1 from me at least :)

piranna:
> Do you want that I cherry-pick this change and set it on the prebuild-install repo?
> +1 from me at least :)

[Done](https://github.com/mafintosh/prebuild-install/pull/2) :-)

piranna:
Any update on this? Should we close it and recommend people to use `prebuild-install` instead?

PR #116, opened 2016-04-30
Add option for custom libc

corbinu:
This is meant to fix https://github.com/mafintosh/prebuild/issues/108

It adds an option for libc argument and LIBC environment variable for use in systems like Alpine linux that use a custom libc.

I was unable to run the test suite on my Mac as it seems it is not currently working for node 6.0 but I am happy to add some tests as advised. 

I have tested it using a scoped package which can be downloaded from NPM as @corbinu/prebuild. 

I also released a version using it and the generated artifacts and travis script can be seen here: https://github.com/corbinu/couchnode


ralphtheninja:
Sweet!

@mafintosh new minor, right?

ralphtheninja:
4.2.0

PR #121, opened 2016-05-16
fix for require test

ralphtheninja:
Closes https://github.com/mafintosh/prebuild/issues/120

@juliangruber I kept the require test where it is, this way we can publish it as a patch and be sure we're not breaking the api


test/download-test.js
```diff
@@ -273,6 +274,28 @@ test('error during download should fail with no dangling temp file', function (t
...
 })
 
+test('should fail if abi is system abi with invalid binary', function (t) {
+  var opts = getOpts()
+  opts.abi = rc.abi
+  opts.pkg.binary = {host: 'http://localhost:8890'}
+
+  var server = http.createServer(function (req, res) {
+    res.statusCode = 200
+    var archive = path.join(__dirname, 'invalid.tar.gz')
```
ralphtheninja: `invalid.tar.gz` just contains a text file renamed to `invalid.node` that should fail when it's required

test/download-test.js
```diff
@@ -273,6 +274,28 @@ test('error during download should fail with no dangling temp file', function (t
   })
 })
 
+test('should fail if abi is system abi with invalid binary', function (t) {
+  var opts = getOpts()
+  opts.abi = rc.abi
```
ralphtheninja: Here's the link to `rc.abi` which should cause a failing test, since it's set in the next commit.

juliangruber:
+1! Great work!

ralphtheninja:
4.2.1

PR #124, opened 2016-05-28
replace simple-get to send requests through configured proxy

csulok:
I've taken a crack at solving https://github.com/mafintosh/prebuild/issues/35. In the comments you hinted at using [got](https://www.npmjs.com/package/got), but I used [request](https://www.npmjs.com/package/request) here, because afaict got doesn't actually read the npm proxy settings.

Couldn't get 1 test to work, where an error event is emitted directly into the response, so this is not perfect. Also, there's remaining work - to create tests with a proxy. Would appreciate any help with that!

### DIFF
--- a/download.js
+++ b/download.js
@@ -41,9 +41,8 @@ function downloadPrebuild (opts, cb) {
         }
 
         fs.mkdir(util.prebuildCache(), function () {
-
-          var stream = fs.createWriteStream(tempFile);
-          stream.on('open', function() {
+          var stream = fs.createWriteStream(tempFile)
+          stream.on('open', function () {
             request({
               url: downloadUrl,
               encoding: null,
@@ -64,13 +63,10 @@ function downloadPrebuild (opts, cb) {
               })
             }).pipe(stream)
           })
-          
-          stream.on('error', function(err) {
+          stream.on('error', function (err) {
             return onerror(err)
           })
-
         })
-        
       })
 
       function onerror (err) {
--- a/download.js
+++ b/download.js
@@ -47,9 +47,8 @@ function downloadPrebuild (opts, cb) {
               url: downloadUrl,
               encoding: null,
               timeout: 3 * 1000
-            }).on('error', function (err) {
-              return onerror(err)
-            }).on('response', function (res) {
+            }).on('error', onerror)
+            .on('response', function (res) {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -23,6 +23,7 @@
     "expand-template": "^1.0.0",
     "ghreleases": "^1.0.2",
     "github-from-package": "0.0.0",
+    "got": "^5.6.0",
```
ralphtheninja: Is this a leftover?

download.js
```diff
@@ -40,27 +40,37 @@ function downloadPrebuild (opts, cb) {
...
+                  if (err) return cb(err)
+                  log.info('renaming to @', cachedPrebuild)
+                  unpack()
+                })
               })
-            })
+            }).pipe(stream)
+          })
+          
+          stream.on('error', function(err) {
```
ralphtheninja: How about `stream.on('error', onerror)`?

download.js
```diff
@@ -40,27 +40,37 @@ function downloadPrebuild (opts, cb) {
...
-                unpack()
+        fs.mkdir(util.prebuildCache(), function () {
+
+          var stream = fs.createWriteStream(tempFile);
+          stream.on('open', function() {
+            request({
+              url: downloadUrl,
+              encoding: null,
+              timeout: 3 * 1000
+            }).on('error', function (err) {
```
ralphtheninja: `.on('error', onerror)`

download.js
ralphtheninja: Why do you do `stream.on('open')` before making the request?

ralphtheninja: @juliangruber @mafintosh  Mind taking a look a the pipe logic here? :)

csulok: stream.on('open') before making the request, because fs.createWriteStream is not immediately available. http://stackoverflow.com/questions/12906694/fs-createwritestream-does-not-immediately-create-file

package.json
csulok: yes, something I looked over. taken care of in upcoming commit

mafintosh:
just wanna mention that we used to use request but switched to simple-get to reduce install time.

csulok:
In that case I understand how got might be a better choice. I'll look at it and see if I can make a new PR later.

PR #127, opened 2016-06-12
Prerelease

### LINKED ISSUE #126: Pre-releases
Mark a release as a pre-release if the version field has one or it's defined a `--prerelease` flag.

piranna:
This fixes #126 and also allow to add extra fields to a currently defined prerelease version.


upload.js
```diff
@@ -15,12 +18,20 @@ function upload (opts, cb) {
     })
   }
 
+  var semver = SemVer(pkg.version)
+  var prerelease = opts.prerelease
+  if (prerelease && prerelease !== true) semver.prerelease.push(prerelease)
```
juliangruber: why should it be not `true`?

piranna: When false (the default) or falsy value, the release is only marked as
prerelease if the package.json version field has it defined; if true or
truthty vamue, it force the release to be a prerelease; and if a string or
a number, it also append its value to the chain of the prerelease field.
Maybe should I add it in the docs or in a comment? Should I also add a
check to allow zero and disallow empty string?

piranna:
Any update on this? Could it be merged?

piranna:
By the way, it's correctly working to generate the [NodeOS prebuilds](https://github.com/NodeOS/NodeOS/releases).

juliangruber:
I haven't gotten back immediately because I was hoping for feedback from other maintainers. I don't have _the_ solution in my mind yet, and I don't want to publish this prematurely and then needing to do a major bump when we want to change it.

This feature is obviously important so we should get it right. I am mostly worried about the workflow with this. With npm prereleases, you strictly only need to publish with a tag, no need for a special `.version` with something like `-beta2` in it. Therefor I was thinking we should do the same, so people don't need to learn one more prerelease workflow.

If we want to go down this route, we also need to add `.tag` as an option for identifying a prebuild. Yes, you could argue that if we just used `.version` for this, we wouldn't need to make this addition. To me though, mirroring npm is more important here.

I might have had other concerns and ideas but can't remember them right now. What does everyone think?

piranna:
> I haven't gotten back immediately because I was hoping for feedback from other maintainers. I don't have the solution in my mind yet, and I don't want to publish this prematurely and then needing to do a major bump when we want to change it.

I agree, it makes sense to be quiet if you think this is so important, just wanted to get some feedback about its progress :-)


> With npm prereleases, you strictly only need to publish with a tag, no need for a special .version with something like -beta2 in it. Therefor I was thinking we should do the same, so people don't need to learn one more prerelease workflow.

I don't fully understand this fragment... Are you talking about having several dit-separated fields in the `prerelease`? That's valid for semver and accepted by npm, in fact each fragment is procesed separately as an array and give precedence to the longer ones according to the spec, so it works as expected.


> If we want to go down this route, we also need to add .tag as an option for identifying a prebuild.

This is not needed since the `prebuild` is identified as unique all on its own, taking in account all the dot-separated items.


> Yes, you could argue that if we just used .version for this, we wouldn't need to make this addition. To me though, mirroring npm is more important here.

I agree on that, but for example `async.js` is publishing the prereleases "as is", without tags. Other thing would be if you want that they don't get downloaded by default when trying to fetch the latest version, but I think this is outside the scope here and maybe mostly it's a bug on `npm` itself (prereleases should be installed *only* when using an explicit version).

juliangruber:
> I don't fully understand this fragment... Are you talking about having several dit-separated fields in the prerelease? That's valid for semver and accepted by npm, in fact each fragment is procesed separately as an array and give precedence to the longer ones according to the spec, so it works as expected.

What is `dit`? With npm you can publish a `3.0.0` as a prerelease if you want, and that's totally valid.

piranna:
> What is dit?

I mean dot-separated ;-)


> With npm you can publish a 3.0.0 as a prerelease if you want, and that's totally valid.

Are you talking about https://docs.npmjs.com/cli/dist-tag? Seems to be more related to distribution channels and similar things, not about prereleases in the semver or Github meaning...

ralphtheninja:
I'm +-0 on this. My main concern is documentation and maybe add an example of how this can be used.

piranna:
> I'm +-0 on this. My main concern is documentation and maybe add an example of how this can be used.

Touchée. I'm going to write them and update the pull-request.

piranna:
I totally forgot about that, sorry. I've added some documentation about the `prerelease` argument and option. Could you be able to review if it's ok? Also, should I rebase on top of the current master? Seems there are no conflicts...

piranna:
Any update on that? Could be reviewed and merged?

piranna:
ping?

ralphtheninja:
@piranna I'm really busy the coming days but I'll try to take a look asap.

piranna:
> @piranna I'm really busy the coming days but I'll try to take a look asap.

Thanks, take your time :-)

vweevers:
I'm -1 on this in its current form. It adds two features:

1. Mark a GitHub release as a prerelease. This is good to have.
2. Modify the version, on which I'm -1. As @juliangruber pointed out this introduces a new workflow, and IMO `package.json` should be the source of truth for `version`).

Closing because this is stale.

I'll leave the issue (#126) open. A new PR is welcome, ideally only tackling feature 1, which I think no one is opposed to and would get a quick merge.

lovell:
Alternative, simpler PR for this at https://github.com/prebuild/prebuild/pull/310 with the `prerelease` option but without any version number modifications.

PR #133, opened 2016-10-25
Fix regression in the strip implementation

oleavr:
The internal `spawn()` utility function's API recently changed, but the `strip()` implementation did not follow along. This fixes the regression.

Sadly this PR introduces a new dependency in order to do proper quoting -- though I'm not familiar enough with prebuild to say whether we can assume that quoting will never be needed, i.e. the `file` argument never contains spaces.

/cc @ralphtheninja

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "os-homedir": "^1.0.1",
     "pump": "^1.0.0",
     "rc": "^1.0.3",
+    "shell-quote": "^1.6.1",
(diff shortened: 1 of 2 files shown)


mafintosh:
thanks for reporting this. i ended up with a little different fix for this in master as i rather not get into argument escaping to not risk potential security bugs.

oleavr:
Sounds great. Thanks for addressing this so quickly! ❤️

PR #134, opened 2016-10-28
Bump node versions and add node 7.0.0

lgeiger:
So `prebuilt --all` will build for Node.js 7 as well.


lgeiger:
@ralphtheninja It looks like https://github.com/ralphtheninja/a-native-module needs to be updated in order to run the test suite on Node 7.

ralphtheninja:
@lgeiger I'll get on that today!

ralphtheninja:
4.4.0

I'll work on the tests once leveldown prebuilts have been made.

@lgeiger thanks!

ralphtheninja:
@lgeiger Aaah you're using prebuild for zmq? :heart:

lgeiger:
Yes we just moved our [repo](https://github.com/zeromq/zeromq.js) into the zeromq org.

It's working really great end to end with Travis and AppVeyor! Thanks a lot for this awesome package.

The only downside is that we still have to build from source if we want to use it inside with electron.

ralphtheninja:
Nice! So happy that this module can contribute to the zmq community. Building for electron is still an enigma for me.

rgbkrk:
Yeah it's ok though. Our fallback with Electron is so clean now that I'm pleased. I'm even happier that we made zeromq easier to install for node users in general. 😄

PR #135, opened 2016-11-24
Build from source if used inside electron

lgeiger:
Since we don't support prebuilts for Electron we should build from source if the `--runtime=electron` flag is used by `npm`.
It is necessary to build native modules from source to ensure that they are build against the correct Electron headers: http://electron.atom.io/docs/tutorial/using-native-node-modules/

This is especially useful if the module is used inside a Atom package: https://github.com/atom/apm/pull/638

### DIFF
--- a/rc.js
+++ b/rc.js
@@ -25,6 +25,12 @@ for (var j = 0; j < npmconfigs.length; ++j) {
   }
 }
 
+// Ensure that modules used inside electron are build from source
+if (process.env['npm_config_runtime'] === 'electron' &&
+    process.argv.indexOf('--build-from-source') === -1) {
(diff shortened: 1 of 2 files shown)


lgeiger:
Could we get this merged and shipped?
We are using `prebuild` in [Hydrogen](https://github.com/nteract/hydrogen/) and the installation is controlled by Atom so we can't use the `--build-from-source` flag.

ralphtheninja:
@juliangruber new minor?

lgeiger:
Thanks @ralphtheninja

ralphtheninja:
4.5.0

@lgeiger No. Thank you! I'm just pushing buttons here :D

PR #136, opened 2016-11-27
remove 0.10, it's dead

ralphtheninja:
The other updates aren't really necessary since it's the major we care about (the abi has changed) but just syncing with what's out there at this point in time

### DIFF
--- a/rc.js
+++ b/rc.js
@@ -25,6 +25,12 @@ for (var j = 0; j < npmconfigs.length; ++j) {
   }
 }
 
+// Ensure that modules used inside electron are build from source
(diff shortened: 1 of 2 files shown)


ralphtheninja:
@juliangruber I want to release https://github.com/mafintosh/prebuild/pull/135 today. Can we merge this as well?

juliangruber:
+1 from me, however this is a breaking change so maybe a +1 from @mafintosh would be great as well

ralphtheninja:
Agree. Well then I can leave it for now and release 4.5.0.

mafintosh:
I'd like to keep this in there as long as 0.10 is default on debian

PR #137, opened 2016-11-29
remove 0.10

ralphtheninja:
(updated my master to upstream, so making a new PR from another branch)


ralphtheninja:
Since this is a breaking change (not api wise) this means releasing 5.0.0. We should update dependencies and stuff before releasing it.

Issue #61, opened 2015-12-27
Prebuilds for Electron and Atom


rgbkrk:
Prebuilding across the nodejs versions as listed in [targets](https://github.com/mafintosh/prebuild/blob/master/targets.js) is super easy! Thanks!

How do I handle builds against the electron headers?

ralphtheninja:
> How do I handle builds against the electron headers?

Good question. I've been wanting to add some `--electron` flag (or similar) that would allow you to do that. No one has asked for it until now.

maxogden:
+1, though im curious how it would work at install time for the user. For example, if I publish a electron build to github releases for my module, when the user runs `npm install`, how does their process know to request the electron binary and not their `require('os').platform` since they are running their globally installed node + npm to install dependencies and not using electron's node to install npm dependencies.

rgbkrk:
Last night I was thinking about it in terms of having a given electron app have the propery binary fetch as a postinstall step to go ahead and grab the right version for the electron version. 
For `electron-rebuild`, it shells out to detect the version of electron, assuming electron-prebuilt is a sibling package of the package - https://github.com/electronjs/electron-rebuild/blob/master/src/main.js#L42. For me, that means having electron-prebuilt be a dependency. We don't need an actual rebuild in those cases, but do require fetching the proper binary at that time.

If there's some other way to get that out of the `package.json`, that would be even better.

At least with `node-pre-gyp`, it would be a matter of asking to reinstall for that package alone:

```
node-pre-gyp install --target=v0.33.0 --runtime=electron
```

I'd like the flow to be simpler than that and you're right. When a user `npm install`s the package, it should *just work*. The hard part is knowing which runtime they're targeting.

maxogden:
Hmm yea I think the best way would be to use some special property in package.json that the process spawned by postinstall can use to decide what binary to download.

```JSON
{
  "name": "a-native-module",
  "scripts": {
    "install": "prebuild --download"
  },
  "dependencies": {
    "prebuild": "^2.3.0"
  },
  "prebuild-runtime": "electron"
}
```

Would that work? That way everything else with prebuild stays that same, but if `prebuild-runtime` is declared then it will override the default

rgbkrk:
I certainly like the `prebuild-runtime` key.

[...]

PR #139, opened 2016-12-01
Remove --abi option

lgeiger:
This removes the `--abi` option since it does the same thing as `--target`. The ABI will now be determined based on the target.

This will allow for `npm install --target=7.0.0` and `npm rebuild --target=7.0.0` without needing to specify a ABI. If i recall correctly `--target` is a "official" `npm` option in contrast to `--abi`.

In the process I also changed the usage of `process.version` to `process.versions.node` since `process.version` would return `v7.0.0` instead of `7.0.0`.

This behavior is also prerequisite for #61 to determine the right version when installing modules with `npm install --runtime=electron --target=1.4.10 --dist-url=https://atom.io/download/electron`.

### DIFF
--- a/abi.js
+++ b/abi.js
@@ -1,6 +1,23 @@
+var ver = require('semver')
 var gypinstall = require('./gypinstall')
 var util = require('./util')
 var error = require('./error')
+
+function getAbiFromTarget (target) {
+  if (!target) return process.versions.modules
+  if (!ver.valid(target)) {
+    throw error.noAbi(target)
+  }
+  if (ver.eq(target, process.versions.node)) return process.versions.modules
+  if (ver.satisfies(target, '^7.0.0')) return '51'
+  if (ver.satisfies(target, '^6.0.0')) return '48'
+  if (ver.satisfies(target, '^5.0.0')) return '47'
+  if (ver.satisfies(target, '^4.0.0')) return '46'
+  if (ver.satisfies(target, '^0.12.0')) return '14'
+  if (ver.satisfies(target, '^0.10.0')) return '11'
+
+  throw error.noAbi(target)
+}
 
 function getAbi (opts, version, cb) {
   var log = opts.log
@@ -50,4 +67,5 @@ function getAbi (opts, version, cb) {
   }
 }
 
-module.exports = getAbi
+exports.getAbi = getAbi
+exports.getAbiFromTarget = getAbiFromTarget
--- a/abi.js
+++ b/abi.js
@@ -1,20 +1,16 @@
-var ver = require('semver')
 var gypinstall = require('./gypinstall')
 var util = require('./util')
 var error = require('./error')
 
 function getAbiFromTarget (target) {
   if (!target) return process.versions.modules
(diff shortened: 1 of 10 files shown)


abi.js
```diff
@@ -1,7 +1,24 @@
...
+  if (!ver.valid(target)) {
+    throw error.noAbi(target)
+  }
+  if (ver.eq(target, process.versions.node)) return process.versions.modules
+  if (ver.satisfies(target, '^7.0.0')) return '51'
+  if (ver.satisfies(target, '^6.0.0')) return '48'
+  if (ver.satisfies(target, '^5.0.0')) return '47'
+  if (ver.satisfies(target, '^4.0.0')) return '46'
+  if (ver.satisfies(target, '^0.12.0')) return '14'
+  if (ver.satisfies(target, '^0.10.0')) return '11'
```
juliangruber: small proposal, do

```js
if (/^0\.10\./.test(target)) return '11'
```

to get rid of one dependency

lgeiger: Great idea 👍

abi.js
```diff
@@ -2,6 +2,19 @@ var gypinstall = require('./gypinstall')
...
 
+function getAbiFromTarget (target) {
+  if (!target) return process.versions.modules
+  if (target === process.versions.node) return process.versions.modules
+  if (/^7\.[0-9]+\.[0-9]+/.test(target)) return '51'
+  if (/^6\.[0-9]+\.[0-9]+/.test(target)) return '48'
+  if (/^5\.[0-9]+\.[0-9]+/.test(target)) return '47'
+  if (/^4\.[0-9]+\.[0-9]+/.test(target)) return '46'
+  if (/^0\.12\.[0-9]+/.test(target)) return '14'
+  if (/^0\.10\.[0-9]+/.test(target)) return '11'
```
juliangruber: what about simplifying more to this?

- `/^7\./`
- `/^6\./`
- `/^5\./`
- `/^4\./`
- `/^0\.12\./`
- `/^0\.10\./`

lgeiger: I wanted to be strict to make sure it throws if a invalid target like `--target=7.a.bc` is given. But I'm happy to simplify it.

lgeiger:
Regarding Electron, a bit more work needs to be done. 

I did some research and made a [list of Electron ABIs](https://github.com/lgeiger/electron-abi) similar to Node.js.
I also did some testing on Linux and Windows with native modules built against different versions of Electron. There are some strange things happening on very old Electron versions but for the most part this looks very promising:
https://travis-ci.org/lgeiger/zmq-prebuilt/builds/180257275
https://ci.appveyor.com/project/lgeiger/zmq-prebuilt/build/1.0.185

ralphtheninja:
> I did some research and made a list of Electron ABIs similar to Node.js.

You should make a module out of that

lgeiger:
> You should make a module out of that

Great idea: https://www.npmjs.com/package/electron-abi

lgeiger:
Superseded by #141

PR #141, opened 2016-12-02
Support prebuilts for Electron

### LINKED ISSUE #61: Prebuilds for Electron and Atom

### LINKED ISSUE #128: force compiling on electron abi=48
Because of https://github.com/electron/electron/issues/5851 prebuilds are failing in a terrible way on latest electron (they'll usually load but strange bugs happen).

This is because the electron abi is pretending to be compatible with node when it currently isn't.
Is there anyway we can force compile when someone is trying to install a prebuild for electron to work around this?

lgeiger:
This PR is based on #139 

This will allow to prebuilt native modules against Electron headers 🎉 

If a module is installed via `npm install --runtime=electron --target=1.4.10` or rebuild with `npm install --runtime=electron --target=1.4.10` it will download the correct binary.
This allows the usage of native modules in Electron or Atom (via [`apm`](https://github.com/atom/apm)) without needing to build from source.

Fixes #61
Fixes #128 

/cc @rgbkrk


lgeiger:
For reference: https://github.com/electron/electron/issues/5851

rgbkrk:
Awesome!

targets.js
```diff
@@ -1,8 +1,15 @@
...
+  {runtime: 'node', target: '5.12.0'},
+  {runtime: 'node', target: '6.9.1'},
+  {runtime: 'node', target: '7.0.0'},
+  {runtime: 'electron', target: '0.30.8'},
+  {runtime: 'electron', target: '0.32.3'},
+  {runtime: 'electron', target: '0.35.6'},
+  {runtime: 'electron', target: '1.0.2'},
+  {runtime: 'electron', target: '1.2.8'},
+  {runtime: 'electron', target: '1.3.12'},
+  {runtime: 'electron', target: '1.4.10'}
```
mafintosh: Lets just support >=1 for electron

mafintosh:
This PR is awesome, although it contains a lot of changes. Summoning @juliangruber @ralphtheninja to take a look

README.md
```diff
@@ -143,7 +145,6 @@ prebuild [options]
   --preinstall  -i  script      (run this script before prebuilding)
   --compile     -c              (compile your project using node-gyp)
   --no-compile                  (skip compile fallback when downloading)
-  --abi                         (use provided abi rather than system abi)
```
mafintosh: why is this removed?

abi.js
```diff
@@ -1,53 +1,29 @@
...
+  } else {
+    if (!target) return process.versions.modules
+    if (target === process.versions.node) return process.versions.modules
+    if (/^8\./.test(target)) return '52'
+    if (/^7\./.test(target)) return '51'
+    if (/^6\./.test(target)) return '48'
+    if (/^5\./.test(target)) return '47'
+    if (/^4\./.test(target)) return '46'
+    if (/^0\.12\./.test(target)) return '14'
+    if (/^0\.10\./.test(target)) return '11'
```
mafintosh: this will require a new release of prebuild every time there is a new abi electron/node release. anyway to have it be more dynamic like before?

abi.js
```diff
@@ -1,53 +1,29 @@
...
 
-  function parse (file) {
-    var res = file.match(/#define\s+NODE_MODULE_VERSION\s+\(?(\d+)/)
-    return (res && res[1])
-  }
+  throw error.noAbi(target, runtime)
 }
 
-module.exports = getAbi
+exports.getAbi = getAbi
```
mafintosh: no need to change the api

targets.js
lgeiger: Any particular reason?
I would prefer to support all electron versions since Electron 1.0.0 was released in May 2016 and a lot of tools are still using 0.37.

README.md
lgeiger: That change was part of #139

abi.js
lgeiger: Node usually bumps the ABI with every major version: https://github.com/nodejs/node/blob/master/ROADMAP.md#stability-policy

Electron tends to bump it with every minor version: https://github.com/electron/electron/issues/5851#issuecomment-246920775

We could use this to determine the ABI

lgeiger:
For testing purpose I published a native module built with this branch: https://github.com/lgeiger/zmq-prebuilt/releases/tag/v4.0.0

The prebuilts work (apart from random hickups) excellent with Node.js and Electron:
https://travis-ci.org/lgeiger/zmq-prebuilt-testing/builds/180841930
https://ci.appveyor.com/project/lgeiger/zmq-prebuilt-testing-hctq0/build/1.0.2

abi.js
lgeiger: I could make a separate module with this logic so we can rely on it. This way we can just do a patch release of the external module.
@mafintosh Would you prefer this approach?

targets.js
mafintosh: Just fewer builds. I don't feel strongly about this, so if *really* want to we can keep the old ones around. I'm all for backwards compat.

README.md
mafintosh: Ok 👍

abi.js
mafintosh: I'm fine with an external module if you promise to keep it up to date :) If you have that module be able to return a list of current targets we could move that logic in there as well so our `--all` command becomes even simpler / more static

mafintosh:
Assuming @juliangruber or @ralphtheninja signs off on this as well I'm LGTM once that final `abi` comment gets fixed. Thanks again @lgeiger

mafintosh:
(Reminder that we should squash these commits if/when merging)

abi.js
lgeiger: I created [`node-abi`](https://github.com/lgeiger/node-abi) and added you to `npm` as well.

targets.js
lgeiger: Looking at the CI builds again there are some strange things happening for Electron<=0.35.0.

So it makes sense to only support >= 1.  Electron 0.36.0 to 1.0.2 have the same ABI so we will still be backwards compatible.
I'll change that 👍

test/rc-test.js
```diff
@@ -42,15 +42,18 @@ test('custom config and aliases', function (t) {
...
     t.end()
   })
 })
 
 test('using --all will build for all targets', function (t) {
   var args = [
-    '--prebuild vX.Y.Z',
-    '--prebuild vZ.Y.X',
+    '--prebuild X.Y.Z',
+    '--prebuild Z.Y.X',
```
juliangruber: the leading `v` should still be supported, so I'd leave those in place

lgeiger: Good point! Though this test doesn't really test if the leading `v` is supported or not.
I just checked that `npm rebuild --target=v7.2.0` works. This means we just need to support it in node-abi as well.

lgeiger: The leading `v` is now supported: https://github.com/lgeiger/node-abi/commit/4980cb30e46f246f5ef4b4a01cac426a4a3e3ffe

prebuild.js
```diff
@@ -1,61 +1,59 @@
...
   var buildLog = opts.buildLog || function () {}
+  opts.target = target
+  opts.runtime = runtime
 
-  if (target[0] !== 'v') target = 'v' + target
-  var buildLogMessage = 'Preparing to prebuild ' + pkg.name + '@' + pkg.version + ' for ' + target + ' on ' + opts.platform + '-' + opts.arch + ' using ' + opts.backend
+  var buildLogMessage = 'Preparing to prebuild ' + pkg.name + '@' + pkg.version + ' for ' + runtime + ' ' + target + ' on ' + opts.platform + '-' + opts.arch + ' using ' + opts.backend
   if (opts.libc && opts.libc.length > 0) buildLogMessage += 'using libc ' + opts.libc
   buildLog(buildLogMessage)
-  getAbi(opts, target, function (err, abi) {
```
ralphtheninja: This was nice to get rid of!

ralphtheninja:
Looks really good!

+1 for squashing

Also, it would be really nice with an example in the README for how to use this with electron.

lgeiger:
Thanks for your reviews!

> +1 for squashing

This can be done easily via the GitHub UI during merge.

> Also, it would be really nice with an example in the README for how to use this with electron.

Sure! I [updated the README](https://github.com/mafintosh/prebuild/pull/141/commits/2167e00245ed0b014ce0fbcea89162ed29c69381?short_path=04c6e90#diff-04c6e90faac2675aa89e2176d2eec7d8).

While updating the README I noticed that io.js is still supported. I'll update `node-abi` to support all node versions listed in https://nodejs.org/en/download/releases/  (even those not used by `--all`).

mafintosh:
@juliangruber looking good to merge now? i'd like to get this out \o/

ralphtheninja:
> This can be done easily via the GitHub UI during merge.

Haha I didn't know you could do that with github until now. Cool :)

ralphtheninja:
@juliangruber Merge+squash at will!

juliangruber:
🍾 this is a big one!

juliangruber:
Is this a major release? Or does the new target detection simply render the `--abi` option obsolete in all cases where it was relied on?

lgeiger:
A major release sounds good to me. 🚀 

> Or does the new target detection simply render the `--abi` option obsolete in all cases where it was relied on?

It does. But I think `--abi` could be used without the target option before. This won't work anymore.

mafintosh:
@juliangruber should probably have squashed on merge but no big deal.

I'm gonna major release this now 🎉

mafintosh:
Out in 5.0.0 - Thanks everyone.

juliangruber:
> @juliangruber should probably have squashed on merge but no big deal.

oh, I thought I did that. Meh!

mafintosh:
No worries. I don't care that much.

PR #142, opened 2016-12-06
Cleanup prebuild script

lgeiger:
Main changes:
- Replace `fs.stat` with `fs.existsSync`
- Use `noop-log` in tests

### DIFF
--- a/prebuild.js
+++ b/prebuild.js
@@ -33,7 +33,8 @@ function prebuild (opts, target, runtime, callback) {
         buildLog('Packing ' + filename + ' into ' + tarPath)
         pack(filename, tarPath, function (err) {
           if (err) return cb(err)
-          cb(null)
+          buildLog('Prebuild written to ' + tarPath)
+          cb(null, tarPath)
         })
       }
     ]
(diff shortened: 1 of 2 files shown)


prebuild.js
```diff
@@ -17,45 +17,39 @@ function prebuild (opts, target, runtime, callback) {
...
-      },
-      function (filename, cb) {
-        buildLog('Packing ' + filename + ' into ' + tarPath)
-        pack(filename, tarPath, function (err) {
-          if (err) return cb(err)
-          cb(null)
-        })
-      }
-    ]
+  if (fs.existsSync(tarPath) && !opts.force) {
```
mafintosh: whats the benefit to using existsSync over stat here? function is async anyway

lgeiger: Personally I think it only makes the code more readable.

There are also discussions about why `existsSync` may be better/faster: https://github.com/nodejs/node/pull/8364

mafintosh: I don't think we should start using sync calls in async methods. It's a bad pattern.

mafintosh:
+1 except for the sync call

prebuild.js
lgeiger: 👍 Removed the last commit.

mafintosh:
👍 Thanks! Do you wanna help maintain this in general?

mafintosh:
5.0.1

lgeiger:
>  Do you wanna help maintain this in general?

Sure, I'm happy to help 🎉

prebuild.js
```diff
@@ -48,13 +49,7 @@ function prebuild (opts, target, runtime, callback) {
...
     }
 
-    // TODO if we can move out buildLog() to the caller, we can simply do
-    // async.waterfall(tasks, callback)
-    async.waterfall(tasks, function (err) {
-      if (err) return callback(err)
-      buildLog('Prebuild written to ' + tarPath)
-      callback(null, tarPath)
-    })
+    async.waterfall(tasks, callback)
```
ralphtheninja: Nice!

PR #144, opened 2016-12-09
Remove node-gyp dependency

### LINKED ISSUE #143: Remove node-gyp dependency?
At the moment the package size of `prebuild` is quite large.
This is mainly caused by including `node-gyp` and `node-ninja` as a direct dependency.

Generally it is possible to rebuild native modules by executing `npm rebuild package-name` with the correct arguments. Such a approach would use the bundled `node-gyp` of npm which would significantly reduce the install time of `prebuild` and modules using it.

Am I missing a major downside of purely relying on `npm` to rebuild native modules?

lgeiger:
This PR brings down the size of `node_modules`  from 10.9MB to 1.3MB.

Fixes #143


README.md
```diff
@@ -163,7 +170,6 @@ prebuild [options]
   --compile     -c              (compile your project using node-gyp)
   --no-compile                  (skip compile fallback when downloading)
   --libc                        (use provided libc rather than system default)
-  --backend                     (specify build backend, default is 'node-gyp')
```
juliangruber: this option will have to stay, for ninja support

prebuild.js
```diff
@@ -1,5 +1,5 @@
 var fs = require('fs')
-var async = require('async')
+var waterfall = require('async-waterfall')
```
juliangruber: nice!

README.md
lgeiger: I guess if we want to keep ninja we need to include it as a dependency again since it isn't bundled with npm.
But ninja requires node-gyp so we can't remove it and won't gain any improvement in package size.

lgeiger:
The node-ninja build fails on Node 0.12. I don't think this is a problem of `prebuild`.
I'm going to disable the integration test on Node 0.12.

.npmignore
```diff
@@ -0,0 +1,3 @@
+node_modules
+npm-debug.log
+test/
```
ralphtheninja: :+1:

test/native-module/package.json
```diff
@@ -4,6 +4,10 @@
   "main": "./index.js",
   "scripts": {
     "install": "../../bin.js --install",
-    "prebuild": "../../bin.js --prebuild 1.4.11 --runtime electron"
+    "prebuild": "../../bin.js --prebuild 1.4.11 --runtime electron",
+    "prebuild:ninja": "../../bin.js --prebuild 7.2.1 --backend node-ninja"
+  },
+  "dependencies": {
+    "node-ninja": "^1.0.2"
   }
```
ralphtheninja: Awesome!

lgeiger:
I discovered `prebuild-install` in #148 which will fix my package size concerns.
Hence this PR isn't essential for me anymore. That is to say: If you prefer to keep `node-gyp` as a direct dependency, I'm happy to extract the integration test and the other minor improvements into a separate PR.

ralphtheninja:
@lgeiger I'm cleaning up a bit. Since `prebuild-install` was created to address this issue (i.e. big dependency tree) I'm closing it.

PR #145, opened 2016-12-09
Add integration test for building native modules

lgeiger:
Add a simple native module and test that it can be built.


juliangruber:
thanks!

PR #150, opened 2016-12-12
Don't default to compile if installed via yarn

### LINKED ISSUE #132: prebuild --install defaults to compile on Yarn
Yarn: https://code.facebook.com/posts/1840075619545360

If you install a native module with Yarn instead of NPM, `prebuild --install` in the module's install script will always default to compiling rather than downloading a suitable binary.

cc @mafintosh

lgeiger:
Fixes #132 

This is the way [`electron-builder`](https://github.com/electron-userland/electron-builder) determines if it's installed via Yarn.

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -3,6 +3,7 @@
 var path = require('path')
 var log = require('npmlog')
 var fs = require('fs')
+var home = require('os-homedir')
 var async = require('async')
 var extend = require('xtend')
 
@@ -12,6 +13,7 @@ var download = require('./download')
 var prebuild = require('./prebuild')
 var build = require('./build')
 var upload = require('./upload')
+var util = require('./util')
 
 var prebuildVersion = require('./package.json').version
 if (rc.version) {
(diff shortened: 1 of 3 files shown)


ralphtheninja:
Seems `.startsWith()` is not available in older versions of node

lgeiger:
@ralphtheninja Thanks for the hint!

ralphtheninja:
@lgeiger New patch?

lgeiger:
@ralphtheninja Sounds good!

ralphtheninja:
5.0.2

bin.js
```diff
@@ -46,8 +48,13 @@ log.info('begin', 'Prebuild version', prebuildVersion)
 delete process.env.NVM_IOJS_ORG_MIRROR
 delete process.env.NVM_NODEJS_ORG_MIRROR
 
+var execPath = process.env.npm_execpath || process.env.NPM_CLI_JS
+
 if (rc.install) {
-  if (!(typeof pkg._from === 'string')) {
+  if (util.isYarnPath(execPath) && /node_modules/.test(home())) {
```
lgeiger: I just tested `v5.0.2` and it still compiles from source.

Stupid me I used `process.cwd()` for testing and added `home()` here. I'll make another PR. Sorry for the noise.

ralphtheninja: Np.

util.js
```diff
@@ -111,6 +111,10 @@ function releaseFolder (opts, version) {
   }
 }
 
+function isYarnPath (execPath) {
+  return execPath ? /^yarn/.test(path.basename(execPath)) : false
+}
```
rgbkrk: :smile:

bin.js
```diff
@@ -46,8 +48,13 @@ log.info('begin', 'Prebuild version', prebuildVersion)
...
+    // From yarn repository
+    rc.download = rc.install
```
rgbkrk: 💯

PR #154, opened 2016-12-15
add test for rc and external config

ralphtheninja:
Simple sanity check tests


ralphtheninja:
@juliangruber I guess we could use some docs as well?

juliangruber:
nice! yeah let's do that

Issue #69, opened 2016-01-15
Cross compiling for ARM/MIPS


reconbot:
Hi,
I'm working with the [tessel project](http://tessel.io) getting precompiled binary modules built for our microcontroller. We have a processes that works with packages that use `node-gyp` and `node-pre-gyp`. It sets up the environment variables and toolchains to build for the tessel (which uses OpenWRT). You can see our proof of concept [here](https://github.com/tessel/t2-compiler/blob/master/compile.sh).

An [issue was raised](https://github.com/tessel/t2-compiler/issues/14) around our first `prebuild` package (level-down) and I have no idea how to setup cross compiling to the mips architecture. Any direction would help a lot. Thanks!

reconbot:
My hunch is this is around the environment variables we set to tell `pre-gyp` about the target architecture.

ralphtheninja:
@reconbot Some background: `prebuild` is a module attempting to solve almost the same use case as `node-pre-gyp`, with the exception that it can build binaries for *all* abi versions and upload them to [github releases](https://github.com/Level/leveldown/releases/tag/v1.4.3) in one go. It also enables you to download them from github during `npm install`. It uses `node-gyp` to do this. If a prebuilt binary can't be found, it will fallback to building the binaries using `node-gyp`.

If I understand you correctly, the problem here boils down to being able to compile `leveldown` for the mips architecture in the first place. I've never tried that myself so I have no knowledge if it's even possible right now.

ralphtheninja:
That said, I'd be happy to help you out nailing this. Having `leveldown` running on `tessel` would be *really* nice!

cc @mafintosh @juliangruber

reconbot:
That's awesome, thank you =)

I think we can break this up.

1) Levldown might not work, that we can figure out separately. I'm reading
it is possible to build for mips, but the current official releases don't
include it.
2) My mess of env vars don't tell `prebuild` to build for mips. I'm either
not setting something or it's getting overridden.

A little later tonight I'm going to try to build a different prebuild
package for the tessel and see how that goes. Maybe it will shed a little
bit more light on things.

-Francis

PS You don't need a tessel to build packages for the tessel. That repo has
a vagrant vm.

---
Francis Gulotta
wizard@roborooter.com

ralphtheninja:
Cool. Let me know what I can do to help.

piranna:
I have a similar problem, but instead to generate i386 versions from an x64 (or viceversa).

[...]

PR #156, opened 2016-12-17
Get target architecture from npm config environment variables

piranna:
This allow to set the architecture for cross-compile environment by using
`npm install --arch=<arch>`.

This should fix issue #69.

### DIFF
--- a/rc.js
+++ b/rc.js
@@ -30,7 +30,7 @@ module.exports = function (pkg) {
   var rc = require('rc')('prebuild', {
     target: pkgConf.target || process.versions.node,
     runtime: pkgConf.runtime || 'node',
-    arch: pkgConf.arch || process.arch,
+    arch: pkgConf.arch || process.env.npm_config_arch || process.arch,
     libc: process.env.LIBC,
     platform: process.platform,
     all: false,


piranna:
> It should be enough to add arch here

You are right, I didn't see that before, but I think doing it that way not only it's overcomplicated but also there's a problem that's already happening with the current npm configs and it's that the subsequent code it's overwritting the arguments given in the command line, making them useless. I believe that array should be removed, or at least the append code should do some checks first. What do you think?

piranna:
I have just checked this patch with a fork of `fuse-bindings` using it and it worked:

```sh
> fuse-bindings@2.10.1 install /home/piranna/Proyectos/NodeOS/node_modules/nodeos-initramfs/build/i686/lib/node_modules/exclfs/node_modules/fuse-bindings
> prebuild --install

prebuild info begin Prebuild version 5.1.2
prebuild info looking for local prebuild @ prebuilds/fuse-bindings-v2.10.1-node-v48-linux-ia32.tar.gz
prebuild info looking for cached prebuild @ /home/piranna/.npm/_prebuilds/https-github.com-mafintosh-fuse-bindings-releases-download-v2.10.1-fuse-bindings-v2.10.1-node-v48-linux-ia32.tar.gz
prebuild http request GET https://github.com/mafintosh/fuse-bindings/releases/download/v2.10.1/fuse-bindings-v2.10.1-node-v48-linux-ia32.tar.gz
prebuild http 404 https://github.com/mafintosh/fuse-bindings/releases/download/v2.10.1/fuse-bindings-v2.10.1-node-v48-linux-ia32.tar.gz
prebuild WARN install No prebuilt binaries found (target=6.9.1 runtime=node arch=ia32 platform=linux)
prebuild info install We will now try to compile from source.
```

Look at the *WARN* line, the `arch` is correctly set to `ia32` although this is a 64 bits machine.

lgeiger:
> there's a problem that's already happening with the current npm configs and it's that the subsequent code it's overwritting the arguments given in the command line, making them useless

I don't think that's the case.
The `rc` module parses `process.argv`. [Here](https://github.com/mafintosh/prebuild/blob/master/rc.js#L31-L43) only the default value is set, hence the command line options take precendence.

piranna:
> I don't think that's the case.
> The rc module parses process.argv. Here only the default value is set, hence the command line options take precendence.

Problem is that the code before is adding the attributes to `process.argv` from the `npm_config_*` environment variables without removing the already defined command line arguments, and seeing the code of `minimist` seems that when an argument is defined several times [it's converted on an Array](https://github.com/substack/minimist/blob/4cf45a26b9af5f4ddab63107f4007485e576cfd3/index.js#L83-L85). I don't know what's doing `rc` in that case, but seems it's not doing too much magic with it so it will be passed directly to the `node-gyp` arguments and probably break... :-/ I definitely think this can be simplified and less error prone by not doing so much magic by using instead the method I have done here.

lgeiger:
Downloading is now handled by [`prebuild-install`](https://github.com/mafintosh/prebuild-install): https://github.com/mafintosh/prebuild-install/pull/21

PR #160, opened 2016-12-30
Remove install and download related functionality

### LINKED ISSUE #148: Use `prebuild-install` as dependency
Since `prebuild-install` package is a stripped-down version of `prebuild` just to download and install the images, use it as dependency instead of having two codes for the download step that would became divergent in the future and lead to problems by having two different versions.

lgeiger:
As discussed in #148, this will remove all installation and download related functionality since it is now provided by [`prebuild-install`](https://github.com/mafintosh/prebuild-install).

The other major change is the removal of the `--prebuild` (`-b`) option since it would provide the same functionality as `--target`.

Fixes #148


README.md
```diff
@@ -55,59 +54,10 @@ If you don't want to use the token on cli you can put it in `~/.prebuildrc`:
...
-
-```
-$ prebuild --install https://example.com/{name}-{version}-{abi}-{platform}{libc}-{arch}.tar.gz
-```
-
-`--install` will download binaries when installing from npm and compile in other cases. If you want `prebuild` to always download binaries you can use `--download` instead of `--install`. Either way, if downloading fails for any reason, it will fallback to compiling the code.
-
-There's also support for `node-pre-gyp` style by utilizing the `binary` property in `package.json`.
-
 ### Formatting urls
```
ralphtheninja: Maybe the section `Formatting urls` is no longer needed now?

ralphtheninja:
LGTM

@mafintosh @juliangruber what do you think?

I guess this would mean a new major :)

README.md
lgeiger: 👍

fanatid:
I tested `prebuild-install` and looks like this package not compile addon from sources if prebuild release not found... is should work in this way?

mathiask88:
@fanatid How did you use it? Did you see the Usage info [here](https://github.com/mafintosh/prebuild-install)?

fanatid:
damn it, I skipped `|| node-gyp rebuild`
@mathiask88 thank you, all works!

ralphtheninja:
6.0.0

piranna:
Now it's time to update all my projects :-P

PR #162, opened 2017-01-31
support abi number as target parameter too

juliangruber:
This is useful so when you pass --runtime=electron --target=50 to prebuild it still works. I need this functionality for prebuild-ci.

In the future this probably will be supported via https://github.com/lgeiger/node-abi/pull/5.


juliangruber:
this was taken care of in node-abi, no change necessary here :)

PR #164, opened 2017-02-02
fix: get target from --target=abi

juliangruber:
otherwise node-gyp would fail

### DIFF
--- a/prebuild.js
+++ b/prebuild.js
@@ -1,6 +1,7 @@
 var fs = require('fs')
 var async = require('async')
 var getAbi = require('node-abi').getAbi
+var getTarget = require('node-abi').getTarget
 var getTarPath = require('./util').getTarPath
 var build = require('./build')
 var strip = require('./strip')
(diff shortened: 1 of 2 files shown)


prebuild.js
```diff
@@ -15,6 +16,7 @@ function prebuild (opts, target, runtime, callback) {
   var buildLogMessage = 'Preparing to prebuild ' + pkg.name + '@' + pkg.version + ' for ' + runtime + ' ' + target + ' on ' + opts.platform + '-' + opts.arch + ' using ' + opts.backend
   if (opts.libc && opts.libc.length > 0) buildLogMessage += 'using libc ' + opts.libc
   buildLog(buildLogMessage)
+  target = getTarget(target, runtime)
```
lgeiger: Maybe we should add a comment why we have to do this.

juliangruber: yeah good idea!

juliangruber:
6.0.2

PR #175, opened 2017-06-13
Allow strip program to be overridden

oleavr:
Through the `STRIP` environment variable. This is sometimes necessary when
cross-compiling, and is consistent with how gyp allows e.g. `CXX` to be set
to override the C++ compiler used. The name `STRIP` was chosen as autotools
also uses it to override its auto-detection.

Note: I wasn't sure if this should be integrated in the `rc` module or done locally,
so I went for the latter. Please let me know what you'd prefer.

### DIFF
--- a/strip.js
+++ b/strip.js
@@ -4,7 +4,7 @@ function strip (file, cb) {
   // TODO no support on windows, noop
   var platform = util.platform()
   if (platform === 'win32') return process.nextTick(cb)
-  util.spawn('strip', stripArgs(platform, file), cb)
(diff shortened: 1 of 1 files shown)


ralphtheninja:
LGTM

juliangruber:
LGTM too, shall we just merge or add docs as well? Either would be fine to me

ralphtheninja:
I can document it

ralphtheninja:
https://github.com/mafintosh/prebuild/commit/df195f08e568437dfad03d6af1469a2bac9f0747

6.2.0

PR #178, opened 2017-07-08
set gyp.devDir so headers are downloaded correctly

ralphtheninja:
Closes https://github.com/mafintosh/prebuild/issues/172

### DIFF
--- a/test/gypbuild-test.js
+++ b/test/gypbuild-test.js
@@ -2,9 +2,10 @@ var test = require('tape')
 var runGyp = require('../gypbuild')
 var util = require('../util')
 var noop = require('noop-logger')
+var osenv = require('osenv')
 
 test('gyp is invoked with correct arguments, release mode', function (t) {
-  t.plan(6)
+  t.plan(7)
   var opts = {
     pkg: {binary: {
       module_name: 'module_name',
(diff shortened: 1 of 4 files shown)


test/gypbuild-test.js
```diff
@@ -43,6 +44,8 @@ test('gyp is invoked with correct arguments, release mode', function (t) {
     log: noop
   }
   runGyp(opts, 'x.y.z', function (err) {
+    var devDir = opts.gyp.devDir
+    t.equal(devDir.indexOf(osenv.home()), 0, devDir + ' should start with home folder')
```
ralphtheninja: It's a little bit silly sometimes when you write test code that's basically the same as the code you wrote.

ralphtheninja:
cc @mathiask88

mathiask88:
Maybe we should give the user the opportunity to pass a `--devdir` argument accordingly to `node-gyp`. Otherwise LGTM.

ralphtheninja:
I hear you, at the same time I want to add as little complexity as possible until someone really needs to configure `~/.node-gyp` to something else.

ralphtheninja:
6.2.1

Issue #149, opened 2016-12-11
Add hyphen to `libc`?


piranna:
On the docs it seems that the platform and the libc fields in the download URL are concatenated, shouldn't them to be separated by an hyphen?

Also, how to setup a `libc` agnostic build? Maybe with `*`?

corbinu:
@piranna When I added libc the thought was to not break compatibility. So * is simply no libc in the url. As for the hyphen given that all platform & architecture combinations have a default LIBC the thought was to simply add it to the end of the platform. That way it worked for those who needed it but didn't confuse/complicate things for those who did not. However I would not be opposed to a change.

For usage feel free to take a look at this repo of mine: https://github.com/corbinu/couchnode/releases

piranna:
Good to know you already have builds using `musl` :-) I asked about the hyphen to use it as separator for each field, I understand about your point of all platforms has a default libc but not always it's glib, that'w what's being assumed here. In fact, there are other systems like Alpine or (my) NodeOS that are musl-based, so by not defining the libc would download by default a glibc-based prebuild and it will not work... :-/ I don't disagree being glibc as the default one, but there should be then some way to detect the platform is not glibc-based and do the correct request automatically...

corbinu:
@piranna I am fine if you can find a way to reliably detect a Linux's default LIBC. I looked and had a very hard time finding one. In my case I simply have set LIBC=musl by default in all my Alpine containers. I don't know if prebuilt would consider adding the hyphen a breaking change. However I defiantly think that step one would be to find a way to detect the the LIBC for a particular Linux disto. If you can do that I would be happy to help you with a PR to integrate it into prebuild.

piranna:
> I am fine if you can find a way to reliably detect a Linux's default LIBC

Maybe asking politely directly to it? :-)

```sh
piranna@slimbook-C16B:~/Trabajo/HBP/whole neuron/prueba11$ readlink /lib/ld-linux.so.2 
i386-linux-gnu/ld-2.24.so
```

This is from my Ubuntu machine, is a 64 bits machine but definitely is `glibc`-based, as it shows the `gnu` string in the symbolic link, for a `musl`-bases system it would have been `i386-linux-musl/ld-2.24.so`. Don't know what would be the string in other systems, but probably they would be checked in a similar way.


> I don't know if prebuilt would consider adding the hyphen a breaking change

It could be fixed by doing several requests if the url is not specified, one with the libc and if it fails and the libc is not specified one without it as currently.

[...]

PR #180, opened 2017-07-13
Add detection for non-glibc libc on Linux

lovell:
Hello, here's an initial attempt to integrate *automagic* libc detection via [detect-libc](https://www.npmjs.com/package/detect-libc), as discussed in #149.

It will only set the `libc` option if not already done so via command line or environment variable, and then only when run on a non-glibc Linux to ensure backwards compatibility with the implicit default of glibc.

There were no existing libc related tests so guidance on where best to add these, if any, is most welcome.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
   ],
   "dependencies": {
     "async": "^2.1.4",
+    "detect-libc": "^1.0.3",
     "execspawn": "^1.0.1",
     "ghreleases": "^1.0.2",
     "github-from-package": "0.0.0",
--- a/rc.js
+++ b/rc.js
@@ -1,11 +1,14 @@
 var minimist = require('minimist')
 var targets = require('node-abi').supportedTargets
+var detectLibc = require('detect-libc')
+
+var libc = process.env.LIBC || (detectLibc.isNonGlibcLinux && detectLibc.family) || ''
 
 var rc = require('rc')('prebuild', {
   target: process.versions.node,
   runtime: 'node',
   arch: process.arch,
-  libc: process.env.LIBC,
+  libc: libc,
   platform: process.platform,
   all: false,
   force: false,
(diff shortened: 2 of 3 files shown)


lovell:
CI failed on Node 0.12 due to `detect-libc`'s use of features like `const` and `spawnSync`. Is there a desire for the next release of `prebuild` to continue to support Node versions <4?

ralphtheninja:
Aye, prebuild should be pure es5. 

@mafintosh wanna chip in?

lovell:
Updated with support for Node v0.10+

lovell:
`node-pre-gyp` is now using [detect-libc](https://www.npmjs.com/package/detect-libc) - see https://github.com/mapbox/node-pre-gyp/pull/326

ralphtheninja:
> Updated with support for Node v0.10+

Prebuild stopped working on 0.12 because of node-gyp, so no longer a hard requirement.

I'd like to move the detection to `rc.js` instead, i.e. to somewhere around here

https://github.com/prebuild/prebuild/blob/61a15bdd6bd5301ae5c6c7cea4317783189f6bce/rc.js#L8

Maybe do something like (at the top)

```js
var libc = process.env.LIBC || (autodetection here)
```

and later use `libc` variable to set the configuration value for the `libc` property.

The reason for this is to have the `prebuild` function as unopinionated as possible and just use `opts.libc`.

lovell:
Updated, thanks, plus I added a test for the existing ability to set `--libc` via the command line to ensure that feature continues to work.

ralphtheninja:
Anyone against this change?

ralphtheninja:
Released in `7.2.0`

lovell:
@ralphtheninja Thank you, I've created https://github.com/prebuild/prebuild-install/pull/55 for the related install-time change.

PR #187, opened 2017-09-12
add support for 'format' command line parameter

egirshov:
Pass through 'format' parameter which is handled by node-gyp configure step
(https://github.com/nodejs/node-gyp/blob/master/lib/configure.js#L216)

### DIFF
--- a/gypbuild.js
+++ b/gypbuild.js
@@ -30,6 +30,8 @@ function runGyp (opts, target, cb) {
     }
     if (opts.debug) args.push('--debug')
 
+    if (opts.format) args.push('--', '--format', opts.format)
+
     gyp({
       gyp: opts.gyp,
       backend: opts.backend,
--- a/README.md
+++ b/README.md
@@ -65,6 +65,15 @@ loaded:
 prebuild-install --binary-name main-binary.node
 ```
 
+The build file format is selected automatically by `node-gyp`, however it is possible to specify needed format explicitly with `--format` parameter.
+This is particularly useful if unusual flavor is required, which could be specified in 'format-flavor' form
+(there is no comprehensive list of formats/flavors available so one has to find possible combinations from `node-gyp` source code).
+For example, in order to build using Makefiles but assume Android cross-compilation:
+
+```
+prebuild --format make-android
+```
+
 ## Uploading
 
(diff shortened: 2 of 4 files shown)


ralphtheninja:
@egirshov Sorry for not responding in a timely fashion. Mind rebasing this onto latest master? Lets get this merged!

ralphtheninja:
@egirshov Remind me again, what the `--format` is used for? I'm looking through the `node-gyp` code to figure it out and at a quick glance it seems related to which system you use for building, e.g. `make` or `msvs` for windows.

What is your use case? Just curious.

egirshov:
I noticed that `node-gyp` supports composite formats (i.e. `format-flavor`), see https://github.com/nodejs/node-gyp/blob/master/gyp/pylib/gyp/__init__.py#L62 while by default the format is either `msvs` or `make` (https://github.com/nodejs/node-gyp/blob/master/lib/configure.js#L216).
In particular, I needed to set the format to `make-android` in order to force cross-compilation mode when building leveldown (the pull request is referenced above, I'll rebase it as well as some point).

I guess having the choice of build system might be useful in other case as well.

gypbuild.js
```diff
@@ -28,6 +28,8 @@ function runGyp (opts, target, cb) {
     }
     if (opts.debug) args.push('--debug')
 
+    if (opts.format) args.push('--', '--format', opts.format)
```
ralphtheninja: Why is `--` needed here? Isn't `--format` part of `node-gyp`'s arguments?

ralphtheninja:
@egirshov Cool! It would be nice if you could add some docs on `--format` with a simple example, perhaps mention the `make-android` use case specifically.

gypbuild.js
ralphtheninja: I'm guessing `node-gyp` will take the arguments after `--` and pass them directly to the python code. I just don't see any docs on it.

egirshov: The logic in `parseArgv` is somewhat convoluted (https://github.com/nodejs/node-gyp/blob/master/lib/node-gyp.js#L135-L147), as far as I understood, the regular arguments (those before `--`) are consumed in that function while after `--` are passed to the python code.

egirshov:
updated readme (added example)

ralphtheninja:
7.4.0

PR #189, opened 2017-10-06
Reduce compressed tarball size by ~1%

lovell:
Hello, the change in this PR offers a quick win to very slightly reduce the output compressed tarball size and therefore very slightly improve tarball download and decompression time.

By way of example, here are the before and after sizes for the latest leveldown binaries:

Before:
```
175256 leveldown-v2.0.0-node-v57-darwin-x64.tar.gz
211712 leveldown-v2.0.0-node-v57-linux-x64.tar.gz
211979 leveldown-v2.0.0-node-v57-win32-ia32.tar.gz
247942 leveldown-v2.0.0-node-v57-win32-x64.tar.gz
```

After:
```
174325 leveldown-v2.0.0-node-v57-darwin-x64.tar.gz
208189 leveldown-v2.0.0-node-v57-linux-x64.tar.gz
211251 leveldown-v2.0.0-node-v57-win32-ia32.tar.gz
246791 leveldown-v2.0.0-node-v57-win32-x64.tar.gz
```

### DIFF
--- a/pack.js
+++ b/pack.js
@@ -27,7 +27,7 @@ function pack (filename, tarPath, cb) {
         tarStream.finalize()
       })
 
-      tarStream.pipe(zlib.createGzip()).pipe(ws).on('close', cb)
+      tarStream.pipe(zlib.createGzip({ level: 9 })).pipe(ws).on('close', cb)
     })
   })
 }


juliangruber:
This reduces tarball size, but also increases compute time, right?

I'm curious what trade off would yield the best overall results here.

lovell:
There is a increase in compression compute time, but this is offset by a slight reduction in decompression compute time. Given these tarballs are compress-once, decompress-many, this feels like a good thing.

There don't appear to be many published benchmarks around gzip compression level, but the two I've found are consistent in demonstrating that higher gzip compression levels decompress faster.

https://www.rootusers.com/gzip-vs-bzip2-vs-xz-performance-comparison/
https://tukaani.org/lzma/benchmarks.html

ralphtheninja:
6.2.2

Issue #188, opened 2017-09-28
building native modules with node-gyp completely broken for 0.12?


ralphtheninja:
When releasing `leveldown@1.9.0` I noticed the following error for node 0.12 https://travis-ci.org/Level/leveldown/jobs/280998747

Does this mean that `node-gyp` no longer can build for node 0.12?

juliangruber:
Yes it seems so, unless we pin to an older version of node-gyp. We should also just drop .12

vweevers:
Apparently 0.x was dropped a while ago: https://github.com/nodejs/node-gyp/pull/1212

vweevers:
From a quick browse through [dependents of prebuild](https://github.com/nice-registry/native-modules/tree/master#all_dependents_of_prebuild), this potentially affects:

- https://github.com/couchbase/couchnode
- https://github.com/frida/frida-node
- https://github.com/level/leveldown-hyper
- https://github.com/octalmage/robotjs
- https://github.com/mafintosh/node-sodium-prebuilt
- https://github.com/datproject/rabin
- https://github.com/mathiask88/node-snap7
- https://github.com/wadey/node-microtime

vweevers:
I might have missed some: those that have custom (hard-to-grep) build processes, no travis or appveyor, and/or no `engines.node` in `package.json`.

PR #190, opened 2017-12-04
fix travis badge + update node versions

ralphtheninja:
Resolves https://github.com/prebuild/prebuild/issues/188 https://github.com/prebuild/prebuild/issues/191


ralphtheninja:
Hmm fails on 0.12. Maybe we should remove support for 0.10 and 0.12 now.

mathiask88:
Maybe we should release one last minor that fixes the dependencies to the level they were 0.x compatible and then go for a new major and drop 0.x?

lgeiger:
:+1: for removing support for Node 0.10 and 0.12

ralphtheninja:
@mathiask88 How would we do that? Revert `node-gyp` to an older version?

mathiask88:
@ralphtheninja Yes I meant to fix the dependencies to the latest version they work with 0.x But I don't mind not doing it.

ralphtheninja:
> @ralphtheninja Yes I meant to fix the dependencies to the latest version they work with 0.x But I don't mind not doing it.

Cool. I still think we should release a new major though when dropping 0.x support.

ralphtheninja:
Will release a new major after this PR has been merged.

ralphtheninja:
7.0.0

PR #171, opened 2017-03-30
make distributed files regex configurable AND support multiple .node files

antoinepairet:
This PR is the same as #122 with conflicts resolved. It replaces #170 and solves #169 

Feedback welcome :-)

PR #122, opened 2016-05-21
make distributed files regex configurable

tmpvar:
This allows people to build / distribute platform specific files, rather than only the `*.node` artifact. This is helpful for including things like statically linked archives (`*.a`) or similar platform specific build artifacts.

PR #192, opened 2018-01-15
Enable multiple files to be included in prebuild packages

daviwil:
This PR finishes the work started by @tmpvar in #122 and @antoinepairet in #171.  It accomplishes two goals:

- Allowing multiple `.node` files to be included in a prebuilt binary package
- Allowing the user to specify a regex using `--collect-files-filter` to be used for including other files in the build output path

A corresponding PR for prebuild-install will be sent later.


daviwil:
There's a test failure I need to look into, plus looking into @todbot's issue.

daviwil:
@todbot, can you try again with my branch to see if you can still reproduce your issue?  I just tried it on my Windows machine and couldn't reproduce it.  I'll try it on my MacBook in a moment to see if it shows up there.  I used this command in the `node-hid` repo that I cloned:

```
prebuild -t 7.0.0 --collect-files-filter "\.(node|lib)$"
```

The resulting .tar.gz file had the expected files, `HID.lib`, `HID.node`, and `hidapi.lib`.

@ralphtheninja I think this PR is ready for review!  Can you check the documentation I added to README.md to see if it's sufficient?  Also, do you want to settle on the `--collect-files-filter` parameter name or possibly change to something like `--included-files-filter`?

help.txt
```diff
@@ -13,3 +13,4 @@ prebuild [options]
   --debug                       (set Debug or Release configuration)
   --verbose                     (log verbosely)
   --version                     (print prebuild version and exit)
+  --collect-files-filter        (regex to match files that will be distributed [default: '\\.(node|a)$'])
```
ralphtheninja: `[default: '\\.(node|a)$']` should be `[default: '\\.(node)$']` according to the value in `rc.js`

ralphtheninja:
> @ralphtheninja I think this PR is ready for review! Can you check the documentation I added to README.md to see if it's sufficient? Also, do you want to settle on the --collect-files-filter parameter name or possibly change to something like --included-files-filter?

Naming is always the most difficult thing. I'm thinking maybe we could shorten it a bit? What about just `--include` or something else that is short and to the point.

I'd like to get some input from other maintainers as well.

Other than that, this looks great!

README.md
```diff
@@ -41,6 +41,21 @@ See [`allTargets`](https://github.com/lgeiger/node-abi#usage) for all available
 
 For more options run `prebuild --help`. The prebuilds created are compatible with [node-pre-gyp](https://github.com/mapbox/node-pre-gyp)
 
+If you'd like to include other files with your prebuilds like additional
+`.node` files or other native libraries, you can pass a file-matching regular
+expression to the  `--collect-files-filter` parameter:
```
ralphtheninja: you can pass a file-matching regular expression to the  `--collect-files-filter` parameter

->

you can pass a file-matching regular expression to `--collect-files-filter`:

README.md
```diff
@@ -41,6 +41,21 @@ See [`allTargets`](https://github.com/lgeiger/node-abi#usage) for all available
...
+
+```
+prebuild -t 8.0.0 --collect-files-filter "\.(node|a)$"
+```
+
+Note that if you include multiple `.node` files, you will need to use the
+prebuild-install's '`--binary-name` parameter to indicate which file should be
```
ralphtheninja: Looks like there's a `'` hanging loose there, just before `--binary-name` :)

help.txt
```diff
@@ -7,10 +7,10 @@ prebuild [options]
   --upload-all  -u  [gh-token]  (upload all files from ./prebuilds folder to github)
   --preinstall  -i  script      (run this script before prebuilding)
   --path        -p  path        (make a prebuild here)
+  --collect-files-filter        (regex to match files that will be distributed [default: '\.node$'])
```
ralphtheninja: Good that you moved this up a bit!

test/collect-artifacts-test.js
```diff
@@ -0,0 +1,35 @@
...
+    t.error(err, 'collected files')
+    t.equal(collected.length, 2, 'found two files')
+
+    // ensure the files are always in the same order
+    collected.sort()
+
+    t.deepEqual(collected, [
+      path.join(release, 'test.node'),
+      path.join(release, 'test.out')
+    ])
```
ralphtheninja: :+1:

help.txt
daviwil: Yep, already caught that!

README.md
daviwil: 👍

README.md
daviwil: Whoops, thanks!  Also, what do you think about `--binary-name` as the `prebuild-install` parameter for choosing the right .node file?  Is there a more obvious name?

daviwil:
I definitely like `--include`, but should it be `--include-regex` just to be super clear about what the parameter accepts?  If you think it's not necessary, I'm happy to go with `--include`.

ralphtheninja:
> I definitely like --include, but should it be --include-regex just to be super clear about what the parameter accepts? If you think it's not necessary, I'm happy to go with --include.

I think `--include-regex` is fine as well.

lgeiger:
This looks great! Thanks everybody for the great work! :tada:

I think using either `--include-regex` or `--include-pattern` would be a good choice.

ralphtheninja:
@juliangruber Do you have any input/comments?

juliangruber:
This looks good to me! I didn't sing off on it because you had requested changes still

daviwil:
@lgeiger `--include-pattern` is a great suggestion, but maybe we should save that in case we also want to support `minimatch` [globbing patterns](https://www.npmjs.com/package/minimatch#usage) in the future?

Just pushed an update with the requested fixes. @ralphtheninja, can you check once again?   Before merging this I'd also like to get confirmation on the `prebuild-install --binary-name` parameter that I invented for the new documentation.  Based on what you decide, I'll send a PR to `prebuild-install` later today to add that parameter.

ralphtheninja:
@juliangruber Approved.

@daviwil `--binary-name` is fine imo.

> I'll send a PR to prebuild-install later today to add that parameter.

After this is merged we can hold off with release until `prebuild-install` is also on par, so we can release them in sync.

ralphtheninja:
Screw holding off the release. This is backwards compatible. Releasing it as `7.1.0`.

daviwil:
Awesome, thanks a lot! The prebuild-install PR will be out as soon as I write a test :)

ralphtheninja:
> Awesome, thanks a lot! The prebuild-install PR will be out as soon as I write a test :)

Thank you! :grinning:

lgeiger:
Excellent thanks for getting this shipped.

PR #194, opened 2018-01-18
downgrade nsp for node 4

ralphtheninja:
This is only devDependencies so no biggie. I tried to get a PR through to `nsp` to remove stupid destructuring issue, but it was rejected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
   },
   "devDependencies": {
     "a-native-module": "^1.0.0",
-    "nsp": "^3.1.0",
+    "nsp": "^2.8.1",
     "rimraf": "^2.4.2",
     "standard": "^10.0.3",
     "tape": "^4.0.1"


juliangruber:
I'm +-0, `nsp` is a good dependency to have up to date, but I can see people being stuck on node 4 too

juliangruber:
If I have to decide, my vote is for future instead of past

ralphtheninja:
We have to live with that tests will not work on node 4 then.

ungoldman:
Might be worth requesting nsp maintain compatibility with Node 4 bit longer.

ralphtheninja:
> Might be worth requesting nsp maintain compatibility with Node 4 bit longer.

@ungoldman I have, but it was rejected.

ralphtheninja:
It's cool though, this only means we can't run the tests on node 4, so no biggie.

vweevers:
How about skipping the `nsp check` command on node 4?

ralphtheninja:
@vweevers Wanna make a PR? :)

vweevers:
#197

PR #197, opened 2018-01-20
Skip nsp on node 4 and 5

vweevers:
Alternative for #194.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,19 @@ env:
     - CXX: g++-4.9
     - CC: gcc-4.9
 
-node_js:
-  - '9'
-  - '8'
-  - '7'
-  - '6'
-  - '5'
-  - '4'
+matrix:
+  include:
+  - node_js: '9'
+    env: [NSP=true]
+  - node_js: '8'
+    env: [NSP=true]
+  - node_js: '7'
+    env: [NSP=true]
+  - node_js: '6'
+    env: [NSP=true]
+  - node_js: '5'
+  - node_js: '4'
+
+script:
+  - npm run test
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -12,10 +12,19 @@ env:
...
+  - node_js: '7'
+    env: [NSP=true]
+  - node_js: '6'
+    env: [NSP=true]
+  - node_js: '5'
+  - node_js: '4'
+
+script:
+  - npm run test
+  - test $NSP && npm run audit
```
vweevers: Need a `|| echo "skip nsp"` at the end.

.travis.yml
```diff
@@ -12,10 +12,19 @@ env:
...
+  - node_js: '7'
+    env: [NSP=true]
+  - node_js: '6'
+    env: [NSP=true]
+  - node_js: '5'
+  - node_js: '4'
+
+script:
+  - npm run test
+  - test $NSP && npm run audit || echo "skip audit"
```
ralphtheninja: Nice one!

vweevers: Hm wait. If `npm run audit` fails, does it fail the build?

ungoldman: @vweevers Ideally it should, right? Looks like operator precedence might short circuit this check.

vweevers:
```
$ test $nope && test $nope || echo "oops"
oops
$ echo $?
0
```

To ensure the exit code of `npm run audit` is the exit code of the line, we can do:

```
if [ "$NSP" ]; then npm run audit; else echo "skip audit"; fi
```

vweevers: #198

PR #203, opened 2018-01-25
Adding support for node-webkit using nw-gyp.

saboya:
This PR adds support to node-webkit runtime.

The changes are pretty straightforward. There's a questionable decision regarding hard-setting `backend` early when `runtime` is set to `node-webkit`, it makes it impossible to use another `backend` when targeting `node-webkit`. I did it because of the log message right after and because it's extremely unlikely that another `backend` is going to be used, but it's technically incorrect.

This PR relies on https://github.com/lgeiger/node-abi/pull/34.
I modified package.json to link directly to my branch so tests would pass for now, so obviously this PR cannot be merged in this state. This branch will be updated accordingly.

Tests are included.

Thanks for your work.


package.json
```diff
@@ -27,11 +27,12 @@
     "github-from-package": "0.0.0",
     "minimist": "^1.1.2",
     "mkdirp": "^0.5.1",
-    "node-abi": "^2.0.0",
+    "node-abi": "git+https://github.com/saboya/node-abi.git#node-webkit",
```
ralphtheninja: This should be a semver version, so it needs to be updated before merge.

test/integration-test.js
```diff
@@ -20,3 +20,18 @@ test('can prebuild a native module for electron', function (t) {
...
+  }
+  rm.sync(path.join(cwd, 'prebuilds'))
+  var file = 'native-v1.0.0-node-webkit-v59-' + process.platform + '-' + process.arch + '.tar.gz'
+  var prebuild = path.join(cwd, 'prebuilds', file)
+  exec('npm run prebuild-node-webkit', { cwd: cwd }, function (error, stdout, stderr) {
+    t.equal(error, null)
+    t.equal(fs.existsSync(prebuild), true)
+    t.end()
+  })
+})
```
ralphtheninja: :+1:

gypbuild.js
```diff
@@ -22,9 +22,14 @@ function runGyp (opts, target, cb) {
     }
     args.push('--target=' + target)
     args.push('--target_arch=' + opts.arch)
-    if (opts.runtime === 'electron') {
-      args.push('--runtime=electron')
-      args.push('--dist-url=https://atom.io/download/electron')
+    switch (opts.runtime) {
```
ralphtheninja: I personally prefer an `if-else` instead of a switch statement. Not sure what others think.

ralphtheninja:
It seems we have been slacking off when it comes to documenting `--runtime`, which needs to be documented for the `electron` and `node-webkit` runtimes. I'll create a separate issue for this.

package.json
saboya: As soon as the PR is accepted by `node-abi`folks, I'll push a fixed version.

test/integration-test.js
```diff
@@ -20,3 +20,18 @@ test('can prebuild a native module for electron', function (t) {
     t.end()
   })
 })
+
+test('can prebuild a native module for node-webkit', function (t) {
+  if (process.versions.modules === '11') {
+    console.log('Skipping test on node 0.10!')
+    return t.end()
+  }
```
ralphtheninja: You can remove this check, since we no longer test on `0.10`. If you feel like it, you can remove the check in the other test as well.

ralphtheninja:
@saboya Mind adding a line in the `README` to show how to build against `node-webkit`? See example for `electron`.

ralphtheninja:
@saboya Also, `help.txt` needs to be updated `[node|electron|node-webkit]`

saboya:
Done!

ralphtheninja:
@saboya Mind rebasing this onto latest master?

package.json
ralphtheninja: @saboya Feel free to update to e.g. `^2.2.0`

saboya:
All set.

ralphtheninja:
@saboya Thanks for your contribution. Would you like to be part of the `prebuild` org?

saboya:
@ralphtheninja Thank you for the invite, would be great :)

Issue #207, opened 2018-01-27
Run tests on AppVeyor


ralphtheninja:

mathiask88:
https://github.com/prebuild/prebuild/pull/208

PR #208, opened 2018-01-28
adds appveyor config

mathiask88:
Closes https://github.com/prebuild/prebuild/issues/207

@ralphtheninja

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,44 @@
+# http://www.appveyor.com/docs/appveyor-yml
+
+# Don't actually build
+build: off
+
+# Skip tag builds
+skip_tags: true
+
+# Set build version format
+version: "{build}"
+
+# Set up build environment
+environment:
+  # Test against these versions of Node.js
+  matrix:
+    - nodejs_version: "9"
+    - nodejs_version: "8"
+    - nodejs_version: "7"
+    - nodejs_version: "6"
+    - nodejs_version: "5"
+    - nodejs_version: "4"
+
+# Build on both platforms
+platform:
+  - x86
+  - x64
+
+# Install scripts (runs after repo cloning)
+install:
+  # Get the latest version of $env:nodejs_version
+  - ps: Update-NodeJsInstallation (Get-NodeJsLatestBuild $env:nodejs_version) $env:platform
+  # Update to latest npm
+  - npm -g install npm 2>&1>nul
+  - SET PATH=%APPDATA%\npm;%PATH%
+  # Output useful info for debugging
+  - node --version
+  - npm --version
+  # Install modules
+  - npm install
+
+# Post-install test scripts
+test_script:
+  # Run module tests
+  - npm test
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,9 +15,7 @@ environment:
   matrix:
     - nodejs_version: "9"
     - nodejs_version: "8"
-    - nodejs_version: "7"
     - nodejs_version: "6"
-    - nodejs_version: "5"
     - nodejs_version: "4"
 
 # Build on both platforms
@@ -29,9 +27,6 @@ platform:
 install:
   # Get the latest version of $env:nodejs_version
   - ps: Update-NodeJsInstallation (Get-NodeJsLatestBuild $env:nodejs_version) $env:platform
-  # Update to latest npm
-  - npm -g install npm 2>&1>nul
-  - SET PATH=%APPDATA%\npm;%PATH%
   # Output useful info for debugging
   - node --version
   - npm --version
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ $ npm install -g prebuild
 ```
 
 [![build status](http://img.shields.io/travis/prebuild/prebuild.svg?style=flat)](http://travis-ci.org/prebuild/prebuild)
+[![Build status](https://ci.appveyor.com/api/projects/status/oy1kk4fpy51net0v/branch/master?svg=true)](https://ci.appveyor.com/project/mathiask88/prebuild)
 [![js-standard-style](https://img.shields.io/badge/code%20style-standard-brightgreen.svg)](http://standardjs.com/)
 
 ## Features
--- a/test/integration-test.js
+++ b/test/integration-test.js
@@ -10,6 +10,8 @@ test('can prebuild a native module for electron', function (t) {
   rm.sync(path.join(cwd, 'prebuilds'))
   var file = 'native-v1.0.0-electron-v50-' + process.platform + '-' + process.arch + '.tar.gz'
   var prebuild = path.join(cwd, 'prebuilds', file)
+  // A quick, temporary fix for a node.js bug (https://github.com/prebuild/prebuild/pull/208#issuecomment-361108755)
+  console.log()
   exec('npm run prebuild', { cwd: cwd }, function (error, stdout, stderr) {
     t.equal(error, null)
     t.equal(fs.existsSync(prebuild), true)
@@ -21,6 +23,8 @@ test('can prebuild a native module for node-webkit', function (t) {
   rm.sync(path.join(cwd, 'prebuilds'))
   var file = 'native-v1.0.0-node-webkit-v59-' + process.platform + '-' + process.arch + '.tar.gz'
   var prebuild = path.join(cwd, 'prebuilds', file)
+  // A quick, temporary fix for a node.js bug (https://github.com/prebuild/prebuild/pull/208#issuecomment-361108755)
+  console.log()
   exec('npm run prebuild-node-webkit', { cwd: cwd }, function (error, stdout, stderr) {
     t.equal(error, null)
     t.equal(fs.existsSync(prebuild), true)


ralphtheninja:
Hmm interesting. We actually got a build error there for node 4 and 32 bit.

mathiask88:
Maybe a singularity, lets wait for the merge build.

ralphtheninja:
Looks like it's actually failing on that particular version. Which is _good_ since we can hopefully reproduce and fix it. (I don't like temporary hickup problems)

vweevers:
I think the problem is in the environment variables. Somewhere, all variables with a value of `undefined` are converted to the *string* `'undefined'` and then npm tries to `require('undefined')` as an [onload-script](https://docs.npmjs.com/misc/config#onload-script).

vweevers:
Able to reproduce on Node.js 4.5.0 32 bits.

mathiask88:
@vweevers Yes, you are right. Let's open an issue for that. But this PR should be fine I think?

vweevers:
> Let's open an issue for that. But this PR should be fine I think?

I can only reproduce with latest npm. So can you remove `npm -g install npm` from `appveyor.yml`? Let's see what happens. I think that's safer anyway, to rely on the npm version bundled with node.

Note we may need latest npm on node 5 to fix https://github.com/nodejs/node-gyp/issues/972. See [leveldown/apppveyor.yml](https://github.com/Level/leveldown/blob/beaf17a2333a8d83d8b5df061bd748277b971ab9/appveyor.yml#L20) as an example.

ralphtheninja:
Hmm, ok. So we're not doing anything wrong? Where is the issue?

emilbayes:
I would also remove node.js version 5 and 7, since they're both EOL 😄 No need to be wasteful with free resources

ralphtheninja:
@emilbayes No reason. Make a PR? :)

appveyor.yml
```diff
@@ -29,9 +27,6 @@ platform:
 install:
   # Get the latest version of $env:nodejs_version
   - ps: Update-NodeJsInstallation (Get-NodeJsLatestBuild $env:nodejs_version) $env:platform
-  # Update to latest npm
-  - npm -g install npm 2>&1>nul
-  - SET PATH=%APPDATA%\npm;%PATH%
```
ralphtheninja: 👍

vweevers: 🤞

ralphtheninja:
Hmm, same probem 😠

vweevers:
There must be some random factor. Locally, I failed to reproduce with a bundled npm version, but let me see what happens if I run the test many many times.

vweevers:
> Locally, I failed to reproduce with a bundled npm version, but let me see what happens if I run the test many many times.

No luck. And I cannot make sense of it. As a hail mary, we could change this:

https://github.com/prebuild/prebuild/blob/d820ab5daf576dea8579c654100329295d7576e1/test/integration-test.js#L13

To:

```js
exec('npm run prebuild', { cwd: cwd, env: {} }, function (error, stdout, stderr) {
```

So that the child process doesn't inherit the env.

ralphtheninja:
> As a hail mary

😆

vweevers:
Pff now everything is failing:

```
Error: Failed to replace env in config: ${APPDATA}
    at C:\Program Files (x86)\nodejs\node_modules\npm\lib\config\core.js:417:13
    at String.replace (<anonymous>)
    at envReplace (C:\Program Files (x86)\nodejs\node_modules\npm\lib\config\core.js:413:12)
    at parseField (C:\Program Files (x86)\nodejs\node_modules\npm\lib\config\core.js:391:7)
    at C:\Program Files (x86)\nodejs\node_modules\npm\lib\config\core.js:334:17
    at Array.forEach (<anonymous>)
    at Conf.add (C:\Program Files (x86)\nodejs\node_modules\npm\lib\config\core.js:333:23)
    at ConfigChain.addString (C:\Program Files (x86)\nodejs\node_modules\npm\node_modules\config-chain\index.js:244:8)
    at Conf.<anonymous> (C:\Program Files (x86)\nodejs\node_modules\npm\lib\config\core.js:321:10)
    at C:\Program Files (x86)\nodejs\node_modules\npm\node_modules\graceful-fs\graceful-fs.js:78:16
    at FSReqWrap.readFileAfterClose [as oncomplete] (fs.js:528:3)
TypeError: Cannot read property 'get' of undefined
    at errorHandler (C:\Program Files (x86)\nodejs\node_modules\npm\lib\utils\error-handler.js:205:18)
    at C:\Program Files (x86)\nodejs\node_modules\npm\bin\npm-cli.js:83:20
    at cb (C:\Program Files (x86)\nodejs\node_modules\npm\lib\npm.js:224:22)
    at C:\Program Files (x86)\nodejs\node_modules\npm\lib\npm.js:262:24
    at C:\Program Files (x86)\nodejs\node_modules\npm\lib\config\core.js:81:7
    at Array.forEach (<anonymous>)
    at Conf.<anonymous> (C:\Program Files (x86)\nodejs\node_modules\npm\lib\config\core.js:80:13)
    at Conf.f (C:\Program Files (x86)\nodejs\node_modules\npm\node_modules\once\once.js:25:25)
    at Conf.emit (events.js:160:13)
    at Conf.add (C:\Program Files (x86)\nodejs\node_modules\npm\lib\config\core.js:337:10)
    at ConfigChain.addString (C:\Program Files (x86)\nodejs\node_modules\npm\node_modules\config-chain\index.js:244:8)
    at Conf.<anonymous> (C:\Program Files (x86)\nodejs\node_modules\npm\lib\config\core.js:321:10)
    at C:\Program Files (x86)\nodejs\node_modules\npm\node_modules\graceful-fs\graceful-fs.js:78:16
    at FSReqWrap.readFileAfterClose [as oncomplete] (fs.js:528:3)
```

vweevers:
I'm guess Appveyor's `npmrc` has a value with `${APPDATA}` in it, expecting `APPDATA` to be defined in the environment. Sigh.

vweevers:
@mathiask88 your initial instinct was right. I'm sorry for leading you down this dark path 😉. I suggest to:

1. Revert bef47cd4eaf8cc280c7da072b14bcf5024f6991d
2. Remove node 4 from `appveyor.yml`
3. Open an issue for it (needs a deep-dive)
4. Merge this PR.

ralphtheninja:
This error really pisses me off because it makes _no_ sense. Why would a 32 bit version of the same node version cause errors related to environment variables?

vweevers:
> Why would a 32 bit version of the same node version cause errors related to environment variables?

It's Windows. I did find https://github.com/nodejs/node/issues/14593, might be related.

mathiask88:
@ralphtheninja Something similar here https://github.com/mapbox/node-pre-gyp/issues/209 ? I just looked over it quickly.

vweevers:
You know what else fixes it for me (with latest npm, because that's the only way I can reproduce)? Merely adding the line:

```js
console.log(process.env.npm_config_onload_script)
```

Which btw, is an empty string. Just bonkers. Some quantum mechanics thing where reading the data changes the data.

mathiask88:
I think this is nothing we will solve today. What about using [this](https://github.com/prebuild/prebuild/blob/52baa195a1e6f11613b7b1352f63affebceec268/test/integration-test.js#L10-L13) dead test code to skip this weird error?
```javascript
  if ((process.versions.modules === '46') && (process.arch === 'ia32')) {
    console.log('Skipping test on node 4.x-ia32!')
    return t.end()
  }
```

Another question is, should we run the `nspcheck` also on windows or is Linux/OSX sufficient?

vweevers:
> Another question is, should we run the nspcheck also on windows or is Linux/OSX sufficient?

Does `nsp` do Windows-specific checks?

ralphtheninja:
Just some info. Tried building on a separate branch on my fork. I've disabled all versions _but_ node 4 32 bit and also all tests _but_ the integration test. If I try to bypass `npm` by doing `exect(`node ../../bin.js...)` I get the same error, so it seems it not related to `npm` fwiw.

https://ci.appveyor.com/project/ralphtheninja/prebuild/build/3

mathiask88:
> Does nsp do Windows-specific checks?

I don't know but I don't think so. At least the code never uses `process.platform`.

@ralphtheninja looks like the [prebuild variable](https://github.com/prebuild/prebuild/blob/52baa195a1e6f11613b7b1352f63affebceec268/test/integration-test.js#L16) closure does not work, because `existsSync(..)` fails with `ENOENT: no such file or directory, open 'C:\projects\prebuild\test\native-module\undefined'` and that is `cwd + undefined variable`

vweevers:
No harm in keeping `nsp`, I think.

ralphtheninja:
Sorry, my bad. It's not the same error. I get an error from `node-gyp` if I bypass `npm`:

```
prebuild ERR! stack Error: ENOENT: no such file or directory, open 'C:\projects\prebuild\test\native-module\undefined'
prebuild ERR! stack     at Error (native)
prebuild ERR! stack     at Object.fs.openSync (fs.js:549:18)
prebuild ERR! stack     at Object.fs.readFileSync (fs.js:397:15)
prebuild ERR! stack     at readCAFile (C:\projects\prebuild\node_modules\node-gyp\lib\install.js:466:15)
prebuild ERR! stack     at download (C:\projects\prebuild\node_modules\node-gyp\lib\install.js:438:22)
prebuild ERR! stack     at C:\projects\prebuild\node_modules\node-gyp\lib\install.js:185:19
prebuild ERR! stack     at C:\projects\prebuild\node_modules\mkdirp\index.js:30:20
prebuild ERR! stack     at FSReqWrap.oncomplete (fs.js:82:15)
prebuild ERR! not ok 
prebuild ERR! build Error: ENOENT: no such file or directory, open 'C:\projects\prebuild\test\native-module\undefined'
prebuild ERR! build     at Error (native)
prebuild ERR! build     at Object.fs.openSync (fs.js:549:18)
prebuild ERR! build     at Object.fs.readFileSync (fs.js:397:15)
prebuild ERR! build     at readCAFile (C:\projects\prebuild\node_modules\node-gyp\lib\install.js:466:15)
prebuild ERR! build     at download (C:\projects\prebuild\node_modules\node-gyp\lib\install.js:438:22)
prebuild ERR! build     at C:\projects\prebuild\node_modules\node-gyp\lib\install.js:185:19
prebuild ERR! build     at C:\projects\prebuild\node_modules\mkdirp\index.js:30:20
prebuild ERR! build     at FSReqWrap.oncomplete (fs.js:82:15)
] killed: false, code: 2, signal: null, cmd: 'node ../../bin.js --target 1.4.11 --runtime electron' }
```

mathiask88:
@ralphtheninja It is the same [error](https://ci.appveyor.com/project/mathiask88/prebuild/build/12/job/e0kxt9lanxx7ixa9#L346) but we have two more errors before from npm

vweevers:
@ralphtheninja @mathiask88 Yea, I think it's the same underlying bug. I've been trying to replicate it sans npm too, using https://github.com/nodejs/node/issues/14593#issuecomment-319602213 as inspiration, but no luck yet.

ralphtheninja:
I tried disabling the code reading `cafile` https://github.com/ralphtheninja/node-gyp/commit/cd2251a8147594a91bb1779343da6ca959dc78b2 and then I can build

ralphtheninja:
Would it be terrible if we skipped running the tests on 32 bit and only on 64 bit?

vweevers:
> Would it be terrible if we skipped running the tests on 32 bit and only on 64 bit?

@mathiask88 suggested to skip only the failing test on 32 bit: https://github.com/prebuild/prebuild/pull/208#issuecomment-361100498, I think that's a good temporary solution.

vweevers:
This is where I got. Minimal repro:

```js
// test.js
var exec = require('child_process').exec

if (process.argv[2] === 'block') {
  // Any blocking operation triggers the bug
  try { require('fs').readFileSync('foobar') } catch (e) {}
}

exec('node -p "process.env.npm_config_onload_script === \'undefined\' ? \'BUG\' : \'OK\'"', function (err, stdout) {
  if (err) throw err
  console.log(stdout)
})
```

````json
"scripts": {
    "test": "node test.js"
}
```` 

Output:

```
> npm test
OK
> npm test -- block
BUG
```

**Comment by ralphtheninja** [comment]
@mathiask88 Yes, lets use your hack to not run on node 4 32 bit. To your other question, I think it's ok to run nsp check on the other platforms for now. Lets just get these tests on windows through first and we can tweak it later.

**Comment by ralphtheninja** [comment]
@mathiask88 Also, if you clean up/rebase this branch onto latest master, you'll get an additional test that will have the same problem. (The `node-webkit` thingie).

**Comment by vweevers** [comment]
It's a node bug. Reproduced without npm:

```js
var exec = require('child_process').exec

process.env.test = ''

if (process.argv[2] === 'trigger') {
  try { require('fs').readFileSync('foobar') } catch (e) {}
}

exec('node -p "process.env.test === \'undefined\' ? \'BUG\' : \'OK\'"', function (err, stdout) {
  if (err) throw err
  console.log(stdout)
})
```

Output:

```
> node test.js
OK
> node test.js trigger
BUG
```

mathiask88:
@vweevers I can repro, but still weird. If you add a `console.log('foo')` before `exec(..)` it works :see_no_evil:

vweevers:
Left a comment on the Node.js thread: https://github.com/nodejs/node/issues/14593#issuecomment-361111588.

Issue #222, opened 2018-07-19
Error in evaluating "build_v8_with_gn" on node-gyp v3.4.0 for Node target v10.x


reconbot:
This is related to https://github.com/prebuild/prebuild/issues/221 and https://github.com/nodejs/node-gyp/issues/1457 but specifically it's a failure to build node 10 binaries on node 8 when using `prebuild --all` The workaround for node-gyp is to pass `--build_v8_with_gn=false` to node-gyp. There isn't a way as far as I can tell to pass that through prebuild.

I'd work around this issue by running `prebuild --all` on node 10 but I hit #221 when I do that. Fixing this issue here is my best bet.

I don't believe there is a problem if we always pass `--build_v8_with_gn=false`

ralphtheninja:
@reconbot Oh, thanks a lot for spending time on this! Need me or someone else to fix this or are you up for making a PR?

reconbot:
I got a new baby at home so it might be a few days before I could take a
crack at it

ralphtheninja:
> I got a new baby at home so it might be a few days before I could take a crack at it

Congrats! :baby: :baby_bottle:

reconbot:
Found time on the train home 🙃

PR #223, opened 2018-07-20
Added fix for building node 10

reconbot:
This variable was added to node 10's build process and is automatically set when building with node 10 but not on older version of node's. Older nodes ignore the flag.

Reference https://github.com/nodejs/node-gyp/issues/1457

Closes https://github.com/prebuild/prebuild/issues/222


ralphtheninja:
> nsp seems to be failing, I believe nsp is going away in favor of NPM itself, isn't it?

That's a good point. I haven't thought about it, but it makes sense I think. Lets get rid of it.

ralphtheninja:
7.6.1

PR #224, opened 2018-07-21
Remove nsp

ralphtheninja:


ralphtheninja:
I'm thinking of releasing a patch first to get the latest changes out and then removing old node versions and release a new major.

PR #227, opened 2018-08-15
fix(rc): don't overwrite user-provided upload token

mbroadst:
This check rewrites the `upload` variable to be a boolean if
upload-all was specified. This will wipe out the user-provided
github auth token if one was indeed provided, and will fail all
subsequent attempts to upload the prebuilds.


mbroadst:
My bad, I misread the help output for the command

Issue #237, opened 2019-02-03
Failing CI


DanielRuf:
https://ci.appveyor.com/project/mathiask88/prebuild/builds/20023810

vweevers:
May have been a temporary failure or fixed on AppVeyor's side in the mean time.

@mathiask88 can you restart that build?

vweevers:
Nevermind. It was not a temporary failure. We should replace `Update-NodeJsInstallation` with `Install-Product node` according to [this thread](https://help.appveyor.com/discussions/problems/18115-nodejs-11-node-is-not-recognized-as-an-internal-or-external-command-operable-program-or-batch-file).

PR #238, opened 2019-03-02
Recursively list files in the build directory when collecting artefacts

zaro:
I have a module where I need to distribute a 3rd party DLL along with the .node library, and it has a specific directory structure that needs to be replicated .

Currently collecting artefacts list files in only one level deep. 
This PR changes the file listing to be recursive, allowing any kind of directory structure to be included in the prebuilt package.

### DIFF
--- a/collect-artifacts.js
+++ b/collect-artifacts.js
@@ -1,10 +1,7 @@
-var fs = require('fs')
-var path = require('path')
 var error = require('./error')
-var recursive = require("recursive-readdir")
+var recursive = require('recursive-readdir')
 
 module.exports = collectArtifacts
-
 
 function collectArtifacts (release, opts, cb) {
   var fileExp = opts['include-regex']
--- a/collect-artifacts.js
+++ b/collect-artifacts.js
@@ -9,7 +9,7 @@ function collectArtifacts (release, opts, cb) {
     if (err) return cb(err)
 
     var collected = files.filter(function filterByRegex (filename) {
-      return fileExp.test(filename)
+      return fileExp.test(filename.substring(release.length + 1))
     })
 
     if (!collected.length) {
--- a/collect-artifacts.js
+++ b/collect-artifacts.js
@@ -1,46 +1,24 @@
 var fs = require('fs')
 var path = require('path')
 var error = require('./error')
+var recursive = require("recursive-readdir")
 
 module.exports = collectArtifacts
 
-function readDirRecursiveSync(directory, relativeDir) {
-  relativeDir = relativeDir || '';
-  var collected
-  var files = fs.readdirSync(directory, {withFileTypes: true})
-  collected = files.map(function filterFiles(dirEnt){
-    if(dirEnt.isFile()) return path.join(relativeDir, dirEnt.name)
-  })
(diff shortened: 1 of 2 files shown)


vweevers:
I suggest using some npm package for the task of reading recursively.

zaro:
Sure, I can change it to use some ready package.  The only reason I did it shit way is I didn't want to introduce additional dependencies.

collect-artifacts.js
```diff
@@ -1,18 +1,15 @@
...
-var path = require('path')
 var error = require('./error')
+var recursive = require('recursive-readdir')
 
 module.exports = collectArtifacts
 
 function collectArtifacts (release, opts, cb) {
   var fileExp = opts['include-regex']
-  fs.readdir(release, function (err, files) {
+  recursive(release, function (err, files) {
```
vweevers: Does `recursive-readdir` yield relative or absolute paths? If relative, we need to keep the `.map()` call below. If absolute, the `fileExp` regular expression will no longer work as expected and we need to fix that.

zaro: It's not absolute,  but it includes the directory already. In this case each filename in `file` in `files` has `release` already pretended. Testing with my setup , the map causes the `release` to be prepended twice.
As for the fileExp I think it should work as expected, especially with the default value which includes '$' .

vweevers: So before this, the regexp would be tested against e.g. a filename `foobar`. Now it will be tested against `release/foobar`. Am I right? I'm not worried about the default value, but about people that use a custom regex.

zaro: Yes, you are right. 
I can't really think of a case where it might be problematic,  but people tend to get really creative with regexes :)
I  guess, I can add something so the regex works on the same portion of the filename.

collect-artifacts.js
```diff
@@ -1,18 +1,15 @@
...
     if (err) return cb(err)
 
     var collected = files.filter(function filterByRegex (filename) {
-      return fileExp.test(filename)
-    }).map(function addPath (filename) {
-      return path.join(release, filename)
+      return fileExp.test(filename.substring(release.length + 1))
```
vweevers: Bear with me.. This will not work in cases where people use a custom release folder (through `binary.module_path` in `package.json`) and it happens to be something (e.g. dots) that `path.join()` (called in `recursive-readdir`) normalizes to a shorter string. That's an edge case, admittedly, but cases like that are why it's generally preferred to use `path` functions rather than "manual" string slicing.

vweevers: Maybe we can use a different package for recursive reading, one that does not prefix the results. Then the filtering and mapping code here can stay the same.

zaro: What do you think about using path.relative(), to get the filename w/o prefix?  Like so :
```js
      return fileExp.test(path.relative(release, filename))
```
It will resolve both to absolute path(according to the doc) before returning the relative part, so it will behave exactly like removing `release`, no matter what kind of path is used for it.

vweevers: Sure. That should work.

vweevers:
Ignore the build failure (that's #237).

vweevers:
Released in 8.2.0. Thanks!

PR #239, opened 2019-03-03
Test recursively collecting artifacts

vweevers:
Follow-up for #238, cc @zaro.


vweevers:
Ignore the failure, that's #237 which is fixed in master.

Issue #246, opened 2019-03-14
strip command output for multiple files incorrect in 8.2.0


dhritzkiv:
As of v8.2.0, the `--strip` option no longer generates a valid strip(1) output:

```sh
strip build/Release/webgl.node,build/Release/obj.target/webgl.node --strip-all
```

which results in `No such file`.

---

Before (in v8.1.2), I believe the output was simply either one of:

```sh
# it may only be stripping the first file
strip build/Release/webgl.node --strip-all
```

or

```sh
# note the space
strip build/Release/webgl.node build/Release/obj.target/webgl.node --strip-all
```

Both of which ultimately run successfully. I'm can't confirm what the actual `strip` command output is, however.

---

I tested v8.1.2 and it works as expected.

This may be the result of the switch to recursive-readdir
https://github.com/prebuild/prebuild/blob/3b8401da1f0363156d4249d9dc7f12436a756aec/collect-artifacts.js#L3

vweevers:
Yes, it's the first case ("it may only be stripping the first file").

> This may be the result of the switch to recursive-readdir

Correct, and apologies for the breaking change. Before 8.2.0, files in nested folders were not included. The default include regex is `\.node$`, which before only matched 1 file in your case, now 2.

There are 2 problems to solve here:

1. Should we not scan nested folders by default? Or tweak the default regex to only include top-level `.node` files? I'm not sure how to to best satisfy everyone's needs.
2. Fix stripping to correctly handle multiple files.

zaro:
about 1. I think either solution is fine. I would personally go for the regex limiting to only top-level files.

lovell:
Hello, I've opened PR #247 to temporarily revert these breaking changes.

vweevers:
@zaro We've reverted the recursive feature in v8.2.1. You're more than welcome to open a new PR - taking the above into account. Thanks!

vweevers:
Thought: what if we used a glob pattern rather than a regex? The default would be `*.node` and if you need nested files you could specify e.g. `**/*.node`.

PR #247, opened 2019-03-18
Revert recursive behaviour to allow non-breaking v8.2.x release

lovell:
Hello, I've just run into the breaking change of #238 and later reported as #246.

The proposed fixes in #246 are still breaking changes.

This PR performs a clean revert of the two commits related to #238 and #239 to allow a non-breaking v8.2.x release without preventing a second attempt in a later breaking v9.0.0 release to include this feature (e.g. by cherry-picking the commits back in again).

### DIFF
--- a/collect-artifacts.js
+++ b/collect-artifacts.js
@@ -1,16 +1,18 @@
+var fs = require('fs')
 var path = require('path')
 var error = require('./error')
-var recursive = require('recursive-readdir')
 
 module.exports = collectArtifacts
 
 function collectArtifacts (release, opts, cb) {
   var fileExp = opts['include-regex']
-  recursive(release, function (err, files) {
+  fs.readdir(release, function (err, files) {
     if (err) return cb(err)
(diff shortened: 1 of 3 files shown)


vweevers:
> The proposed fixes in #246 are still breaking changes

Could you clarify? I'm not opposed to reverting at all, just feel like I'm missing something if the fix you're referring to is to only include top-level files by default. Why would that be breaking?

lovell:
Hi Vincent, thanks for the quick reply.

The proposed fix to modify the default `include-regex` sadly won't help modules that specify a value for it e.g. [sharp](https://github.com/lovell/sharp/blob/master/.prebuildrc).

vweevers:
@lovell Mind if we pull you in for review in a future PR that reintroduces the feaure?

lovell:
@vweevers With pleasure, thank you.

vweevers:
8.2.1

PR #251, opened 2019-07-12
Pin cmake-js to ~5.2.0 to fix building on Windows

vweevers:
Some change between `cmake-js` 5.2.0 and 5.3.0 broke our AppVeyor tests ([logs](https://ci.appveyor.com/project/mathiask88/prebuild/builds/25930825/job/6tciae0pajf6a36r)), as well as locally on my machine. See https://github.com/cmake-js/cmake-js/issues/186.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   ],
   "dependencies": {
     "async": "^2.1.4",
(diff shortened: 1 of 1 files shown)


vweevers:
9.0.1

PR #254, opened 2019-08-03
Unlock cmake-js

vweevers:
Undoes #251, because https://github.com/cmake-js/cmake-js/pull/187 (5.3.1) fixed builds on Windows.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "n-api"
   ],
   "dependencies": {
-    "cmake-js": "~5.2.0",
+    "cmake-js": "^5.3.2",
     "detect-libc": "^1.0.3",
     "each-series-async": "^1.0.1",
     "execspawn": "^1.0.1",
--- a/test/integration-cmake-test.js
+++ b/test/integration-cmake-test.js
@@ -6,46 +6,43 @@ var rm = require('rimraf')
 
 var cwd = path.join(__dirname, 'native-module-cmake')
 
-// See https://github.com/cmake-js/cmake-js/issues/186
-if (process.platform !== 'win32' || process.arch !== 'ia32') {
-  test('can prebuild a cmake-js native module for node', function (t) {
-    rm.sync(path.join(cwd, 'prebuilds'))
-    var file = 'native-v1.0.0-node-v57-' + process.platform + '-' + process.arch + '.tar.gz'
-    var prebuild = path.join(cwd, 'prebuilds', file)
-    // A quick, temporary fix for a node.js bug (https://github.com/prebuild/prebuild/pull/208#issuecomment-361108755)
-    console.log()
-    exec('npm run prebuild', { cwd: cwd }, function (error, stdout, stderr) {
-      t.equal(error, null)
-      t.equal(fs.existsSync(prebuild), true)
-      t.end()
-    })
+test('can prebuild a cmake-js native module for node', function (t) {
+  rm.sync(path.join(cwd, 'prebuilds'))
+  var file = 'native-v1.0.0-node-v57-' + process.platform + '-' + process.arch + '.tar.gz'
+  var prebuild = path.join(cwd, 'prebuilds', file)
+  // A quick, temporary fix for a node.js bug (https://github.com/prebuild/prebuild/pull/208#issuecomment-361108755)
+  console.log()
+  exec('npm run prebuild', { cwd: cwd }, function (error, stdout, stderr) {
+    t.equal(error, null)
+    t.equal(fs.existsSync(prebuild), true)
+    t.end()
   })
+})
 
-  test('can prebuild a cmake-js native module for electron', function (t) {
-    rm.sync(path.join(cwd, 'prebuilds'))
-    var file = 'native-v1.0.0-electron-v50-' + process.platform + '-' + process.arch + '.tar.gz'
-    var prebuild = path.join(cwd, 'prebuilds', file)
-    // A quick, temporary fix for a node.js bug (https://github.com/prebuild/prebuild/pull/208#issuecomment-361108755)
-    console.log()
-    exec('npm run prebuild-electron', { cwd: cwd }, function (error, stdout, stderr) {
-      t.equal(error, null)
-      t.equal(fs.existsSync(prebuild), true)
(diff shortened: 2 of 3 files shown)


vweevers:
I don't use `cmake-js`. Could someone (who has opinions about https://github.com/cmake-js/cmake-js/issues/186#issuecomment-518150342 and https://github.com/cmake-js/cmake-js/issues/188) take over?

vweevers:
Rebased on master. Note that we temporarily disabled `cmake-js` tests on 32-bit Windows (https://github.com/prebuild/prebuild/pull/255#issuecomment-520122761). This PR shouldn't be merged though until 32-bit builds are fixed in `cmake-js`.

Julusian:
Has this been tested in a while? Is it still broken with cmake-js 6? Was this when running 32bit windows, or producing a 32bit build from 64bit windows?
I have taken on some maintenance of cmake-js, and am willing to figure out the problem if there is still one.

Also, why is cmake-js a dependency rather than devDependency?
Anyone using cmake-js will still need to make it a dependency in their project to make sure it is available for users where no prebuild is available, so cant you rely on that and find that version to execute?

vweevers:
> Has this been tested in a while?

No, there was no one here to help. Good to have you

> Is it still broken with cmake-js 6?

I don't know. Relevant: https://github.com/prebuild/prebuild/pull/291 bumps `cmake-js`.

> Was this when running 32bit windows, or producing a 32bit build from 64bit windows?

When producing a 32bit build from 32bit Node.js (on either 32bit or 64bit Windows).

> I have taken on some maintenance of cmake-js, and am willing to figure out the problem if there is still one.

The original problem is that `cmake-js` (at least 5.x) requires setting `"arch": "ia32"` in the addon's `package.json` (which is a problem if you want to support both 64 and 32bit) rather than defaulting to `process.arch`. Although `cmake-js` does take an `--arch` flag; I don't remember why that didn't work. We set that here:

https://github.com/prebuild/prebuild/blob/290cf8bb4581759206d01aef3709c3c5af1940c0/cmakebuild.js#L10

> Also, why is cmake-js a dependency rather than devDependency?

It's part of our semver contract. Making it a devDependency would suggest we support other `cmake-js` versions.

Julusian:
> When producing a 32bit build from 32bit Node.js (on either 32bit or 64bit Windows).

OK, Ill give this a test.
I am pretty sure that the arch property now works (according to the cmake output with 7.0.0-1, not tried to run the resulting binary). I did find it didn't work for macos, but I have a fix for that. It is possible it may fall into the same issue if running x64 node on arm64 os, I should test that.
As for it not defaulting to process.arch, I would guess that is cmake-js not passing a value and the compiler defaulting to x64. But I dont see harm in explicitly forcing the arch.

> It's part of our semver contract. Making it a devDependency would suggest we support other cmake-js versions

Being an optional peerDependency might work better then? (https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta)

As it stands, I did not realise for over a year that prebuild had cmake-js as a dependency. So I have had 6.x as a dependency in my package. Meaning that when I invoke cmake-js it will be using 6.x, but when I invoke prebuild it will be using the 5.x that you have as a dependency. Which could lead to weird issues. At the very least, it isnt clear that I should have been using 5.x the whole time, its only possible to discover by looking at the dependencies in your package.json.

As an optional peerDependency, it will be very clear that I need to bring a version of cmake-js (it will fail to run otherwise), and if the one I choose is not deemed compatible, then I will get a warning during install.

vweevers:
Peer dependencies often create more problems than they solve. Once we unlock and bump `cmake-js` we can keep it updated going forward, so there won't be a problem anymore.

Julusian:
I've given 32bit on windows a test, and it looks to be working fine to me with cmake-js 6.3.0.
I took a fresh windows 11 vm, installed 32bit node14 and let it install the build tools (it chose vs2017). 
Using a clone of https://github.com/julusian/node-jpeg-turbo/tree/v1.1.2 (the tag is important as newer master is testing unreleased builds of cmake-js), I can do a `npm install` and `npm test` as expected, so by itself cmake-js is picking the correct architecture.

I also tried using a 64bit version of node to target 32bit, and it did work with `--arch=ia32` (not x86), running the tests again with the 32bit version of node (also the cmake log did show it looking correct). I dont see why that would be the cause of anything here, but it is simple to add as an alias and might avoid some confusion.

I repeated this all with node18 (which chose vs2019). vs2022 worked too after figuring out how to install it, and applying an unreleased patch from master

Can you think of a scenario that I have missed?

vweevers:
Nice, thanks. Sounds like we can remove this if-statement, to re-enable 32bit tests:

https://github.com/prebuild/prebuild/blob/290cf8bb4581759206d01aef3709c3c5af1940c0/test/integration-cmake-test.js#L10

Julusian:
Make sure to update cmake-js to v6 too, rather than the v5 you are currently referencing

Julusian:
> Peer dependencies often create more problems than they solve. 

If `prebuild` was supposed to be installed as a dependency then I would agree with you, but as it is supposed to be used a dev-dependency (and so will only ever be at the top level), that removes any scenario I can think of where there could be a versioning issue.  

---------

I returned to this as today I released 7.0.0 of cmake-js. In this version are some breaking changes in how users must write their cmake files, which means that prebuild will likely be unable to produce their prebuilds. The cli is unchanged, so if `cmake-js` was a peer dependency, you could support it with a semver minor change rather than a semver major, which risks causing other users the opposite problem.

That said, I have switched away from `prebuild` everywhere I was using it, so this is no longer a problem for me

vweevers:
> If `prebuild` was supposed to be installed as a dependency then I would agree with you, but as it is supposed to be used a dev-dependency (and so will only ever be at the top level), that removes any scenario I can think of where there could be a versioning issue.

Good point.

> That said, I have switched away from `prebuild` everywhere I was using it, so this is no longer a problem for me

👍 I don't use either of these projects (anymore) so I'm gonna close this PR. If anyone wants to use latest `cmake-js` then a PR is welcome to make it a peer dependency as @Julusian suggested.

Issue #253, opened 2019-07-24
Prebuild N-API module using cmake-js


s1hofmann:
**prebuild version**: 9.0.1
**cmake-js version**: 5.3.1
**node version**: 10.16.0
**platform**: macOS 10.14.6

Hi all,

I'm trying to prebuild an N-API module using cmake-js.
Following the readme, my configuration looks like this:

> package.json
```
...
"scripts": {
...
    "install": "prebuild-install --verbose || cmake-js rebuild",
    "prebuild": "prebuild --backend cmake-js -t 3 -r napi",
...
},
"binary": {
    "napi_versions": [
        3,
        4
    ]
}
...
```

`cmake-js rebuild` works fine, but when trying to prebuild, `cmake-js` runs into problems.
```
info TOOL Using Unix Makefiles generator.
ERR! OMG Unknown runtime: napi
ERR! OMG Unknown runtime: napi
```

`prebuild` tries to run `cmake-js` with the following parameters:

```
cmake-js rebuild --runtime-version=3 --arch=x64 --runtime=napi
```

Problem is, `cmake-js` does not know about runtime `napi` or runtime-version `3`:

```
Available settings:

runtime: application's target runtime, possible values are:
    node: Node.js
    nw: nw.js
    electron: Electron
runtimeVersion: version of the application's target runtime, for example: 0.12.1
arch: architecture of application's target runtime (eg: x64, ia32, arm). Notice: on non-Windows systems the C++ toolset's architecture's gonna be used despite this setting. If you don't specify this on Windows, then architecture of the main node/io.js runtime is gonna be used, so you have to choose a matching nw.js runtime.
```

Am I missing out on something? Any help or hints appreciated! 😃 

P.S. Thanks for your work here! We're relying on `prebuild` and `prebuild-install` in our projects and it helps us a lot! Keep up the great work! 👍

vweevers:
Building as a regular node addon (`-r node`) might work. I don't know how `cmake-js` handles N-API versioning if at all.

@oleavr @jschlight @NickNaso Any advice as to what needs to happen here? Thanks!

jschlight:
@s1hofmann is your project publicly available?

jschlight:
It looks like `cmakebuild.js` needs to be updated to support N-API like `gypbuild.js`. I'm reestablishing my development environment and will continue to assess this.

s1hofmann:
[...]

PR #255, opened 2019-08-09
cmake-js support for N-API add-ons

jschlight:
This PR adds support for using `cmake-js` as a backend for building N-API add-ons.  

See #253.


jschlight:
It looks like the 32-bit Windows builds on AppVeyor are failing. I do not have access to a 32 bit Windows machine so I'm hoping someone with experience with the AppVeyor builds can help me understand why they are failing.

vweevers:
@jschlight 32-bit on Windows is a bit wonky atm. It seems `cmake-js` desires `arch="ia32"` to be set in `package.json`, in order to select the right toolset. I would prefer it to be determined by the bitness of node, see https://github.com/cmake-js/cmake-js/issues/186#issuecomment-520122177.

Maybe add an if-statement to the relevant tests, to skip 32-bit until that discussion is settled.

jschlight:
Thank you @vweevers. I disabled the CMake `ia32` unit tests and the builds and tests are now passing.

README.md
```diff
@@ -146,14 +146,20 @@ Native modules that are designed to work with [N-API](https://nodejs.org/api/n-a
...
+```
 
-In the absense of a need to compile against a specific N-API version, the value `3` is a good choice as this is the N-API version that was supported when N-API left experimental status. 
+In the absence of a need to compile against a specific N-API version, the value `3` is a good choice as this is the N-API version that was supported when N-API left experimental status. 
 
 Modules that are built against a specific N-API version will continue to operate indefinitely, even as later versions of N-API are introduced. 
 
 ### Defining the `NAPI_VERSION` Value
 
+The N-API header files supplied with Node use the w`NAPI_VERSION` preprocessor value supplied by the user to configure each build to the specific N-API version for which the native add-on is being built. In addition, the module's C/C++ code can use this value to conditionally compile code based on the N-API version it is being compiled against.
```
vweevers:
```suggestion
The N-API header files supplied with Node use the `NAPI_VERSION` preprocessor value supplied by the user to configure each build to the specific N-API version for which the native addon is being built. In addition, the module's C/C++ code can use this value to conditionally compile code based on the N-API version it is being compiled against.
```

README.md
```diff
@@ -146,14 +146,20 @@ Native modules that are designed to work with [N-API](https://nodejs.org/api/n-a
...
+
+`prebuild` supports two backend build systems: [`node-gyp`](https://github.com/nodejs/node-gyp) and [`cmake-js`](https://github.com/cmake-js/cmake-js). The `NAPI_VERSION` value is configured differently for each build system.
```
vweevers:
```suggestion
`prebuild` supports two build backends: [`node-gyp`](https://github.com/nodejs/node-gyp) and [`cmake-js`](https://github.com/cmake-js/cmake-js). The `NAPI_VERSION` value is configured differently for each backend. 
```

For consistency; the existing [help section](https://github.com/prebuild/prebuild#help) calls them "build backends".

cmakebuild.js
```diff
@@ -6,9 +6,12 @@ function runCmake (opts, target, cb) {
     if (err) return cb(err)
 
     var args = ['rebuild']
-    args.push('--runtime-version=' + target)
+    if (opts.runtime !== 'napi') args.push('--runtime-version=' + target)
     args.push('--arch=' + opts.arch)
-    args.push('--runtime=' + opts.runtime)
+    if (opts.runtime !== 'napi') args.push('--runtime=' + opts.runtime)
+    if (opts.runtime === 'napi' && target) {
+      args.push('--CDnapi_build_version=' + target)
```
vweevers: What happens when `target` isn't defined?

test/integration-cmake-test.js
```diff
@@ -6,6 +6,9 @@ var rm = require('rimraf')
 
 var cwd = path.join(__dirname, 'native-module-cmake')
 
+if (process.arch !== 'ia32') { // see https://github.com/cmake-js/cmake-js/issues/186
```
vweevers:
```suggestion
if (process.platform !== 'win32' || process.arch !== 'ia32') { // see https://github.com/cmake-js/cmake-js/issues/186
```

Doesn't actually make a difference, it's just to communicate intent.

vweevers:
Few nits, one question, LGTM otherwise.

cmakebuild.js
jschlight: If `target` is not defined, the `napi_build_version` CMake customizable setting is not specified to CMake. In this case the `NAPI_VERSION` C preprocessor define value created by CMake is empty and the N-API build defaults to the most current N-API version supported by the Node version being built against.

I've not been able to create a situation where this happens. But I imagine the `prebuild` project could possibly evolve in such a way that it might happen. If so, the code is ready.

vweevers:
9.1.0

PR #256, opened 2019-08-17
Upgrade dependencies and add node 12 to CI

vweevers:
I intend for this to be included in a 9.1.0 release. Upgraded all dependencies except for `node-gyp` (deserves more scrutiny and probably a major bump here) and `cmake-js` (see #254).

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -3,7 +3,7 @@
 var path = require('path')
 var log = require('npmlog')
 var fs = require('fs')
-var async = require('async')
+var eachSeries = require('each-series-async')
 var napi = require('napi-build-utils')
 
 var pkg = require(path.resolve('package.json'))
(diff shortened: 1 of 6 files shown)


bin.js
```diff
@@ -3,7 +3,7 @@
 var path = require('path')
 var log = require('npmlog')
 var fs = require('fs')
-var async = require('async')
+var eachSeries = require('each-series-async')
```
lovell: The similarly-named [async-each-series](https://www.npmjs.com/package/async-each-series) module is significantly more depended upon and appears to provide the same API.

vweevers: I picked `each-series-async` because of its simple code, typically don't care about how many dependents a package has. I'm fine with either though.

vweevers:
9.1.0

PR #261, opened 2019-09-14
Fix node-gyp cache conflict between node and electron

### LINKED ISSUE #249: Building for Electron 5.0.0 instead builds for Node 5.0.0
If this command is run:
```
prebuild -r electron -t 5.0.0
```
It actually builds for Node v5.0.0.  However, this command builds for Electron correctly:
```
prebuild -r electron -t 5.0.6
```
And since the former command seems to be what is run when doing `prebuild --all`, this means that `prebuild` fails for Electron v5.

vweevers:
Adapted from `prebuildify`. Closes #249.

### DIFF
--- a/gypbuild.js
+++ b/gypbuild.js
@@ -33,6 +33,7 @@ function runGyp (opts, target, cb) {
 
   gyp({
     gyp: opts.gyp,
+    runtime: opts.runtime,
     backend: opts.backend,
     log: opts.log,
     args: args,
--- a/gyp.js
+++ b/gyp.js
@@ -1,6 +1,6 @@
 var assert = require('assert')
 var path = require('path')
-var osenv = require('osenv')
+var os = require('os')
 
 var backends = {
   'node-gyp': require('node-gyp')(),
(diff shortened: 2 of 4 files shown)


gypbuild.js
```diff
@@ -29,6 +31,11 @@ function runGyp (opts, target, cb) {
   }
   if (opts.debug) args.push('--debug')
 
+  // Since electron and node are reusing the versions now (fx 6.0.0) and
+  // node-gyp only uses the version to store the dev files, they have started
+  // clashing. To work around this we explicitly set devdir to tmpdir/runtime(/target)
+  args.push('--devdir=' + path.join(os.tmpdir(), 'prebuild', opts.runtime || 'node'))
```
vweevers: Hm, when running tests locally I don't see this directory being created.

vweevers: Ah. Need to modify this instead:

https://github.com/prebuild/prebuild/blob/4ac850fc87bf9baef097701f7f770dffaec548e9/gyp.js#L28

PR #264, opened 2019-12-18
Update node-gyp to 6.x

dwieeb:
I didn't find anyone talking about this in the issues, so maybe this PR will spark some conversation.

I didn't personally run into issues while doing this in my project:

```
cd node_modules/prebuild && npm i node-gyp@5
```

Breaking changes in 5.x: https://github.com/nodejs/node-gyp/blob/master/CHANGELOG.md#v500-2019-06-13
Breaking changes in 4.x: https://github.com/nodejs/node-gyp/blob/master/CHANGELOG.md#v400-2019-04-24

---

Any chance the version of `node-gyp` could be configurable?

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "mkdirp": "^0.5.1",
     "napi-build-utils": "^1.0.1",
     "node-abi": "^2.2.0",
-    "node-gyp": "^3.0.3",
+    "node-gyp": "^5.0.7",
     "node-ninja": "^1.0.1",
     "noop-logger": "^0.1.0",
     "npm-which": "^3.0.1",
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "mkdirp": "^0.5.1",
     "napi-build-utils": "^1.0.1",
     "node-abi": "^2.2.0",
-    "node-gyp": "^5.0.7",
+    "node-gyp": "^6.0.1",
     "node-ninja": "^1.0.1",
     "noop-logger": "^0.1.0",
     "npm-which": "^3.0.1",


vweevers:
> Any chance the version of node-gyp could be configurable?

That's dangerous because we use the programmatic API of `node-gyp`. You can however pass a custom instance: https://github.com/prebuild/prebuild#javascript-api

lovell:
If we're going for a semver major increment, should we bump node-gyp all the way to the latest v6.x to include improvements for handling multiple Python versions?

vweevers:
I didn't realize 5x wasn't the latest! Yes, let's update to 6x.

package.json
```diff
@@ -31,7 +31,7 @@
     "mkdirp": "^0.5.1",
     "napi-build-utils": "^1.0.1",
     "node-abi": "^2.2.0",
-    "node-gyp": "^3.0.3",
+    "node-gyp": "^5.0.7",
```
vweevers: Let's go straight to 6x.

dwieeb: Only reason I didn't go straight to 6.x is that the `latest` dist tag for node-gyp on npm is 5.x.

vweevers: Hm OK. I asked why that is: https://github.com/nodejs/node-gyp/pull/2013#issuecomment-571229235

rvagg: my bad, it was supposed to be `latest`, the only major diff for 6.x is that it will pick `python3` over `python2` where it exists on the system, otherwise the two are in sync, so 👍 to going straight to 6. We also have a possible 7 coming out soon too, but we're still maintaining 5 for as long as npm keeps on using it.

vweevers:
10.0.0

PR #271, opened 2020-08-18
Support namespaced packages in upload-all

### LINKED ISSUE #245: Work with namespaced packages
It doesn't seem like `prebuild --upload-all` works with packages that have namespaced names (i.e., `@cypress/registry-js`. The problem occurs when uploading the built archives to GitHub:

```
prebuild info begin Prebuild version 8.2.0
prebuild info build Uploading 1 prebuilds(s) to Github releases
prebuild ERR! build Error: can only upload files
prebuild ERR! build     at fs.stat (C:\projects\registry-js\node_modules\ghreleases\ghreleases.js:65:23)
prebuild ERR! build     at FSReqWrap.oncomplete (fs.js:153:5)
```

It seems like `prebuild` is trying to upload the entire `prebuilds/@cypress` directory that is generated instead of the individual `.tar.gz` files within it.

pimterry:
A small change to fix #245, based on the [last comment](https://github.com/prebuild/prebuild/issues/245#issuecomment-471728217) there. This adds an extra dependency on glob, but it's already a transitive dependency anyway.

Rather than do just one level (all that's strictly necessary to make namespaced packages work) this is recursive, so it flattens the tree and uploads every file in `./prebuilds`. Happy to debate that, but I think it more accurately matches the documented behaviour of `upload-all`:

> upload all files from ./prebuilds folder to github

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -5,6 +5,7 @@ var log = require('npmlog')
 var fs = require('fs')
 var eachSeries = require('each-series-async')
 var napi = require('napi-build-utils')
+var glob = require('glob')
 
 var pkg = require(path.resolve('package.json'))
 var rc = require('./rc')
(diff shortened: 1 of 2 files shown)


lovell (approved):
LGTM

vweevers:
10.0.1

PR #277, opened 2021-04-27
Cmake is producing a clean build everytime

takameyer:
cmakebuild.js is hardcoded to 'rebuild', which cleans all build artifacts beforhand.  This is problematic for incremental builds in CI.  This hardcodes it to 'build' instead

### DIFF
--- a/cmakebuild.js
+++ b/cmakebuild.js
@@ -5,7 +5,7 @@ function runCmake (opts, target, cb) {
   which('cmake-js', function (err, cmakeJsPath) {
     if (err) return cb(err)
 
-    var args = ['rebuild']
+    var args = ['build']
     if (opts.runtime !== 'napi') args.push('--runtime-version=' + target)
     args.push('--arch=' + opts.arch)
     if (opts.runtime !== 'napi') args.push('--runtime=' + opts.runtime)


vweevers (approved):
I don't use cmake myself so would like to hear from cmake users. I have no objections (except that it's asymmetrical with node-gyp where we do use `rebuild`).

vweevers:
@takameyer To remove the risk of breakage and speed up landing this PR, perhaps the behavior should be opt-in. Via a new command line option, like `--no-clean` or `--no-rebuild`. WDYT?

takameyer:
@vweevers that seems fair.  I would prefer a `--clean` flag, but I think for consistency a `--no-clean` flag would be appropriate.  Are there are places where this could be implemented?

vweevers:
The new option can be added to defaults here:

https://github.com/prebuild/prebuild/blob/7c0810338b36772ba5f1141ae41e22a6d81f43bc/rc.js#L8-L10

Like so:

```js
var rc = require('rc')('prebuild', { 
  clean: true,
  ...
}
```

Which isn't strictly required, but keeps us aware of supported options. No additional code is needed for argument parsing, because our parser ([`minimist`](https://github.com/substack/minimist)) will recognize `--no-clean` as a boolean out of the box. Then to use the new option, add `opts.clean === false ? 'build' : 'rebuild'` to 2 places:

https://github.com/prebuild/prebuild/blob/8dd24c3999dab03950b7a7d3a40d5580d3064b20/cmakebuild.js#L8

https://github.com/prebuild/prebuild/blob/7c0810338b36772ba5f1141ae41e22a6d81f43bc/gypbuild.js#L11

We could also do `opts.clean ? 'rebuild' : 'build'` but that would require updating unit tests & the programmatic API (which don't use `rc.js`).

Lastly, a new unit test would be good to have. For gyp you can copy [this test](https://github.com/prebuild/prebuild/blob/7c0810338b36772ba5f1141ae41e22a6d81f43bc/test/gypbuild-test.js#L55-L74) to a new test in the same file.

PR #280, opened 2021-08-17
Support stripping multiple files

sergiou87:
As I was trying to change [the `--include-regex` parameter to include multiple files](https://github.com/desktop/desktop-trampoline/blob/8b38cd56a192f423b0e89121641f8526b86ed627/package.json#L18-L20) in the prebuild bundle, while preserving the `--strip` parameter, I noticed `strip` wasn't prepared to receive an array of files (which is what `collect` generates, even when it's only 1 file). In the end, the `spawn` function received an args parameter that looked like this: `[ [ 'build/Release/desktop-trampoline' ], '-Sx' ]`

This PR changes `strip` to accept multiple files, and processes those files sequentially.

### DIFF
--- a/prebuild.js
+++ b/prebuild.js
@@ -51,7 +51,7 @@ function prebuild (opts, target, runtime, callback) {
 
     if (opts.strip) {
       tasks.splice(1, 0, function (filenames, cb) {
-        buildLog('Stripping debug information from ' + filenames)
+        buildLog('Stripping debug information from ' + filenames.join(', '))
         strip(filenames, function (err) {
           if (err) return cb(err)
           cb(null, filenames)
--- a/prebuild.js
+++ b/prebuild.js
@@ -50,11 +50,11 @@ function prebuild (opts, target, runtime, callback) {
     ]
 
     if (opts.strip) {
-      tasks.splice(1, 0, function (filename, cb) {
-        buildLog('Stripping debug information from ' + filename)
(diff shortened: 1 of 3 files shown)


prebuild.js
```diff
@@ -50,11 +50,11 @@ function prebuild (opts, target, runtime, callback) {
     ]
 
     if (opts.strip) {
-      tasks.splice(1, 0, function (filename, cb) {
-        buildLog('Stripping debug information from ' + filename)
-        strip(filename, function (err) {
+      tasks.splice(1, 0, function (filenames, cb) {
+        buildLog('Stripping debug information from ' + filenames)
```
vweevers, line 54:
```suggestion
        buildLog('Stripping debug information from ' + filenames.join(', '))
```

vweevers (changes requested):
One nit, LGTM otherwise

prebuild.js
sergiou87: Thanks! Fixed in [81ffea1](https://github.com/prebuild/prebuild/pull/280/commits/81ffea1a70d60d6a818ca026cd001db84dd46291) from your suggestion 😄

vweevers (approved):

lovell (approved):
I've tested locally and all seems to work as expected, thank you very much for the PR.

vweevers:
11.0.3

PR #282, opened 2021-09-14
chore: bump node-abi version to 3.0.0

xueqingxiao:
Closes https://github.com/prebuild/prebuild/issues/281

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,4 +7,11 @@ matrix:
   - node_js: '12'
     os: linux
   - node_js: '10'
-
+    os: linux
+  - node_js: '14'
+    os: osx
+  - node_js: '12'
+    os: osx
+  - node_js: '10'
+    os: osx
+  
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,19 +2,9 @@ language: node_js
 
 matrix:
   include:
+  - node_js: '14'
+    os: linux
   - node_js: '12'
     os: linux
   - node_js: '10'
-    os: linux
-  - node_js: '8'
-    os: linux
-  - node_js: '6'
-    os: linux
-  - node_js: '12'
-    os: osx
-  - node_js: '10'
-    os: osx
-  - node_js: '8'
-    os: osx
-  - node_js: '6'
-    os: osx
+
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,8 +8,7 @@ environment:
   matrix:
     - nodejs_version: "12"
     - nodejs_version: "10"
-    - nodejs_version: "8"
-    - nodejs_version: "6"
+    - nodejs_version: "14"
 
 # Build on both platforms
 platform:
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -2,19 +2,9 @@ language: node_js
...
     os: linux
   - node_js: '12'
-    os: osx
+    os: linux
   - node_js: '10'
-    os: osx
-  - node_js: '8'
-    os: osx
-  - node_js: '6'
-    os: osx
```
ralphtheninja, line 20: Why are you removing `osx`?

ralphtheninja (changes requested):

.travis.yml
xueqingxiao: my bad, I will revert osx support.

ralphtheninja:
It's probably easier to get this PR through if you just bumped the dependencies. Sorry, but to avoid other discussions of removing support for older node.js versions etc.

ralphtheninja:
@vweevers Do we bump major version when no longer supporting older node versions? I forget.

mafintosh:
My personal take on node support + semver is that I normally major bump it unless the node version is very old at this stage, which I'd say 10 is - aka semverish 🤣

xueqingxiao:
@ralphtheninja it's looks good for me, but we should find out that is the better way to resolve broken build with electron 14. which is related to #281.

ralphtheninja:
> @ralphtheninja it's looks good for me, but we should find out that is the better way to resolve broken build with electron 14. which is related to #281.

I understand. If you just updated the dependencies we could release a patch pretty quickly, rather than having to debate removing node versions, which isn't related to what you really want to achieve. And we can postpone that for later. Makes sense?

xueqingxiao:
> And
yep, but node-abi@3.0.0 is required minimal node v10 that is why I'm updated node engine.

xueqingxiao:
> > @ralphtheninja it's looks good for me, but we should find out that is the better way to resolve broken build with electron 14. which is related to #281.
> 
> 
> 
> I understand. If you just updated the dependencies we could release a patch pretty quickly, rather than having to debate removing node versions, which isn't related to what you really want to achieve. And we can postpone that for later. Makes sense?

yep, but node-abi@3.0.0 is required minimal node v10 that is why I'm updated node engine.

ralphtheninja:
> yep, but node-abi@3.0.0 required minimal node version is v10 that is why I'm updated node engine. you can see release note at [here](https://github.com/electron/node-abi/releases/tag/v3.0.0).

Fair enough. It wasn't really clear from the completely empty pull request message :wink:

ralphtheninja (approved):

vweevers:
> Do we bump major version when no longer supporting older node versions? I forget.

IMO yes. We can't see on which node versions prebuild is used. What we do know is that dropping support can be a painful breaking change for consumers. In comparison, for folks on recent node, a semver-major signal is noise but not painful.

vweevers (approved):

xueqingxiao:
@ralphtheninja @vweevers could you guys kindly merge the pr and release a new version?

ralphtheninja:
`11.0.0`

xueqingxiao:
Thanks a lot.

PR #288, opened 2022-01-26
Upgrade to version 2 of detect-libc

lovell:
Hello, this is the "sibling" change for the approved https://github.com/prebuild/prebuild-install/pull/166 and should also be considered `semver-patch` as `prebuild` currently requires Node.js >= 10.

The change in this PR can be published independently of the `prebuild-install` change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   ],
   "dependencies": {
     "cmake-js": "~5.2.0",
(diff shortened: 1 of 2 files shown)


vweevers (approved):

vweevers:
11.0.2

Issue #286, opened 2022-01-14
Update dependencies: node-gyp stills using v6 it is currently in v8


augustnmonteiro:

zdm:
When you will update dependebcies?
It now works with visual studio 2022 on github actions.

zdm:
All you need to do is to change `"node-gyp": "^6.0.1"` to `"node-gyp": "^8.0.0"` in `package.json`.
API interface not changed and it just works.

vweevers:
Two separate PRs are welcome:

1. Replace Travis (and Appveyor) with a GitHub Actions workflow, because our CI is currently broken and we'd like to test all of this. Also replace the badges in the README.
2. Bump `node-gyp`, and also:
   - Go over the programmatic internals of `node-gyp` that we use here, which are not expected to be documented in the changelog of `node-gyp`
   - Update `engines.node` in `package.json` according to the minimum version supported by `node-gyp`
   - Update the Node.js version matrix in the GItHub Actions workflow. We may want to drop Node.js 10 at this time.

verhovsky:
Please update node-gyp. The latest version of prebuild can't build packages for Electron 20:

```
  CXX(target) Release/obj.target/tree_sitter_runtime_binding/src/binding.o
In file included from ../src/binding.cc:1:
/var/folders/n_/mxm2h0s50k3991k3s7kzydxc0000gn/T/prebuild/electron/20.0.0/include/node/node.h:27:2: error: "It looks like you are building this native module without using the right config.gypi.  This normally means that you need to update electron-rebuild (>=3.2.8) or node-gyp (>=8.4.0) if you're building modules directly."
#error "It looks like you are building this native module without using the right config.gypi.  This normally means that you need to update electron-rebuild (>=3.2.8) or node-gyp (>=8.4.0) if you're building modules directly."
 ^
```

> Go over the programmatic internals of node-gyp that we use here, which are not expected to be documented in the changelog of node-gyp

Please just update it instead of making me update it manually with `overrides`.

lovell:
PR for part 2 at https://github.com/prebuild/prebuild/pull/304

lovell:
prebuild v12.0.0 now available with dependency on node-gyp v9.4.0

https://github.com/prebuild/prebuild/blob/1398629bc8debb33d2c3ab62253217d41b0f32f3/package.json#L34

PR #304, opened 2023-07-06
Add support for Node.js 20, drop support for Node.js 10 and 12

lovell:
This is a semver major change, please see https://github.com/prebuild/prebuild/issues/286 for discussion and details.

It also:

- Upgrades all dependencies to their most recent version that still supports Node.js 14.
- Synchronises the `engines.node` property with that of the latest `node-gyp` we're upgrading to.
- Ensures all tests run on all platforms as I noticed some of the cmake-js tests were optionally being skipped.
- Ensures the `native-module-napi-cmake` test includes the CMake config required for use with Node-API.
- Removes the no-longer-required `mkdirp` dependency as Node.js has provided this feature since version 10.
- Removes unused CI "badges" from the readme

Additions based on feedback:

- Defines deployment target in test fixtures (rather than CI config) so tests can be run locally on macOS

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,10 +13,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: [18, 16, 14, 12, 10]
+        node: [20, 18, 16, 14]
         os: [windows-2019, macos-11, ubuntu-20.04]
         arch: [x64]
         include:
+          - os: windows-2019
+            arch: x86
+            node: 20
           - os: windows-2019
             arch: x86
             node: 18
@@ -26,12 +29,6 @@ jobs:
           - os: windows-2019
             arch: x86
             node: 14
-          - os: windows-2019
-            arch: x86
-            node: 12
-          - os: windows-2019
-            arch: x86
-            node: 10
     timeout-minutes: 10
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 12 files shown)


.github/workflows/ci.yml
```diff
@@ -40,8 +37,7 @@ jobs:
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
         with:
-          python-version: "3.10"
-        # See https://github.com/nodejs/node-gyp/issues/1574#issuecomment-554187531
```
verhovsky, line 44: This comment is still relevant and applies to the "Configure macOS env" step below.

lovell: I've moved this to the `binding.gyp` fixture so tests can be run locally on macOS, thanks for highlighting it.

lovell:
As possibly the ultimate test of this PR, the [sharp v0.32.2 prebuilt binaries](https://github.com/lovell/sharp/releases/tag/v0.32.2) were created using this change and everything worked as expected, with no further changes and without any problems :tada:

https://github.com/lovell/sharp/blob/v0.32.2/package.json#L156

lovell:
@JoshuaWise Would you be interested in reviewing this upgrade to the latest version of `node-gyp`? It looks like it would have helped you with https://github.com/WiseLibs/better-sqlite3/pull/870 (too late, sorry) and should allow you to remove the "overrides" workaround it introduced.

sibelius (approved):

verhovsky:
Can this be merged/released please? Currently prebuild doesn't work if you have the latest Python version installed (3.11).

lovell:
Sure, given there are no further comments and I've used the fork upon which this PR is based to create the prebuilt binaries the last 4 releases of sharp without any problems, then I think we're good to merge. A release will involve a follow-up PR for the semver major version bump.

PR #310, opened 2023-09-21
Add prerelease option

lovell:
This is useful if you want to prevent a newly-created release being labelled by GitHub as "latest". The GitHub UI provides a checkbox to change this status later.

This is a simpler version of https://github.com/prebuild/prebuild/pull/127 that doesn't attempt to implement any semver logic, only the `--prerelease` option and associated tests, which will add the `prerelease` property to the underlying [GitHub API call](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release).

Fixes https://github.com/prebuild/prebuild/issues/126


vweevers (approved):
LGTM

lovell:
v12.1.0

PR #311, opened 2023-10-12
Use arch instead of target_arch

gergof:
When building for a different architecture using node-gyp, the arch option should be used. Using target_arch populates the variable usable in the binding.gyp file, but node-gyp will still do a build with the system's architecture.

### DIFF
--- a/gypbuild.js
+++ b/gypbuild.js
@@ -15,7 +15,7 @@ function runGyp (opts, target, cb) {
   } else {
     args.push('--target=' + target)
   }
-  args.push('--target_arch=' + opts.arch)
+  args.push('--arch=' + opts.arch)
   if (opts.runtime === 'electron') {
     args.push('--runtime=electron')
     args.push('--dist-url=https://electronjs.org/headers')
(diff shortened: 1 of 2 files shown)


lovell:
Thanks for the PR. There's some logic in `node-gyp` that suggests `target_arch` is the expected name.

https://github.com/nodejs/node-gyp/blob/b3d41aeb737ddd54cc292f363abc561dcc0a614e/lib/build.js#L59

There's also a known bug in `node-gyp` relating to this - see https://github.com/nodejs/node-gyp/issues/2865 - if that bug was fixed, would this change still be necessary?

gergof:
I think it still would be necessary. 
The target_arch variable is overwritten here: https://github.com/nodejs/node-gyp/blob/b3d41aeb737ddd54cc292f363abc561dcc0a614e/lib/create-config-gypi.js#L69

The documentation is also suggesting using --arch: https://github.com/nodejs/node-gyp/tree/b3d41aeb737ddd54cc292f363abc561dcc0a614e#command-options

This was probably not the case before, but since updating prebuild to version 12, even though target_arch is set to x64, the flags passed to the compiler state that the build arch should be arm64. With this change it works as expected without any side effects.

vweevers:
See also https://github.com/prebuild/prebuildify/pull/53

lovell (changes requested):
Please can you update the failing "gyp is invoked with correct arguments" tests.

gergof:
I've made the changes, the tests are passing now.

lovell (approved):
Thank you, LGTM

PR #312, opened 2023-10-26
Support Node.js 21

verhovsky:
I don't know if there's more that needs to be done besides this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "glob": "^7.2.3",
     "minimist": "^1.2.8",
     "napi-build-utils": "^1.0.2",
-    "node-abi": "^3.47.0",
+    "node-abi": "^3.51.0",
     "node-gyp": "^9.4.0",
     "node-ninja": "^1.0.2",
     "noop-logger": "^0.1.1",


package.json
```diff
@@ -27,10 +27,10 @@
     "execspawn": "^1.0.1",
     "ghreleases": "^3.0.2",
     "github-from-package": "0.0.0",
-    "glob": "^7.2.3",
+    "glob": "^10.3.10",
     "minimist": "^1.2.8",
     "napi-build-utils": "^1.0.2",
-    "node-abi": "^3.47.0",
+    "node-abi": "^3.51.0",
```
verhovsky, line 33: This is the change we care about.

https://github.com/electron/node-abi/commit/85418439eef4e47cb53e180447caa69da2f00e97

lovell:
It looks like existing semver range for `node-abi` includes its latest version, so perhaps you need to update your local lockfile?

verhovsky:
Running

```sh
rm -rf node-modules package-lock.json
npm install
```

in my package's directory installed node-abi 3.52.0 and everything works now, thanks.

PR #315, opened 2024-01-17
Update dependencies

### LINKED ISSUE #313: Upgrade node-gyp for Python 3.12 support
See also https://github.com/nodejs/node/pull/50209

GitHub Actions will switch to that Python version by default sooner or later.

orgads:
Excluding glob and standard, which require code changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "node-api"
   ],
   "dependencies": {
-    "cmake-js": "^7.2.1",
+    "cmake-js": "^7.3.0",
     "detect-libc": "^2.0.2",
     "each-series-async": "^1.0.1",
     "execspawn": "^1.0.1",
@@ -30,7 +30,7 @@
     "glob": "^7.2.3",
     "minimist": "^1.2.8",
     "napi-build-utils": "^1.0.2",
-    "node-abi": "^3.47.0",
+    "node-abi": "^3.54.0",
     "node-gyp": "^10.0.1",
     "node-ninja": "^1.0.2",
     "noop-logger": "^0.1.1",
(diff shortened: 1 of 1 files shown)


orgads:
These were 2 separate commits - one for standard (which required all the const changes) and one for the rest.

I rebased and swapped them.

package.json
```diff
@@ -21,16 +21,16 @@
...
   "dependencies": {
-    "cmake-js": "^7.2.1",
+    "cmake-js": "^7.3.0",
     "detect-libc": "^2.0.2",
     "each-series-async": "^1.0.1",
     "execspawn": "^1.0.1",
     "ghreleases": "^3.0.2",
     "github-from-package": "0.0.0",
-    "glob": "^7.2.3",
+    "glob": "^10.3.10",
```
lovell, line 30: `glob` versions 9+ use a Promise-based API so this upgrade will require a code change (and highlights a possible unit test blind spot).
https://github.com/prebuild/prebuild/blob/f8d82a1d9f1ae17e2a389414d242e22ee8d2efde/bin.js#L45

package.json
```diff
@@ -39,13 +39,13 @@
     "nw-gyp": "^3.6.6",
     "rc": "^1.2.8",
     "run-waterfall": "^1.1.7",
-    "tar-stream": "^3.1.6"
+    "tar-stream": "^3.1.7"
   },
   "devDependencies": {
     "a-native-module": "^1.0.0",
-    "rimraf": "^4.4.1",
+    "rimraf": "^5.0.5",
```
lovell, line 46: Can we remove this dependency entirely by switching to `fs.rmSync(path, { recursive: true })`?

lovell (changes requested):

package.json
orgads: Thanks for spotting that! I moved it to a separate PR (depends on this one though).

I don't know how to test it without actually uploading stuff :)

package.json
orgads: It doesn't work well. See #319

lovell (approved):

PR #316, opened 2024-01-17
Update node-gyp to v10

orgads:
Fixes #313.

PR #317, opened 2024-01-17
Drop support for Node 14

orgads:
Even 16 is already EOL. Needed for upgrading node-gyp.


lovell:
Superseded by https://github.com/prebuild/prebuild/pull/316

